PR #8, opened 2014-12-24
ENH: add pandas.hdfstore suppport

jreback:
```

In [3]: from pandas import DataFrame

In [4]: df = pandas.util.testing.makeMixedDataFrame()

In [5]: uri_hdfstore = 'hdfstore://foo1.h5::/data'

In [6]: uri_pytables = 'pytables://foo2.h5::/data'

In [7]: uri_h5py     = 'h5py://foo3.h5::/data'

In [8]: into(uri_hdfstore,df)
Out[8]: HDFTable [HDFStore]: [path->foo1.h5, datapath->/data]

In [9]: into(uri_pytables,uri_hdfstore)
Out[9]: HDFTable [PyTables]: [path->foo2.h5, datapath->/data]

In [10]: into(uri_h5py,uri_pytables)
Out[10]: HDFTable [h5py]: [path->foo3.h5, datapath->/data]

In [11]: into(DataFrame, uri_hdfstore)
Out[11]: 
   A  B     C          D
0  0  0  foo1 2009-01-01
1  1  1  foo2 2009-01-02
2  2  0  foo3 2009-01-05
3  3  1  foo4 2009-01-06
4  4  0  foo5 2009-01-07

In [12]: into(DataFrame, uri_hdfstore, where='A>3')
Out[12]: 
   A  B     C          D
4  4  0  foo5 2009-01-07

In [14]: drop('foo2.h5')

In [15]: drop('foo3.h5')

In [16]: data = [('Alice', 1), ('Bob', 2), ('Charlie', 3)]

In [17]: into('foo1.h5::/data', data, dshape=dshape("3 * {id: string, amount: int64}"))
Out[17]: HDFTable [HDFStore]: [path->foo1.h5, datapath->/data]

In [18]: drop('foo1.h5')
```
the use of uri's in resource that need a hierarchical disambiguation
based on priority and the raising of ``NotImplementedError`` by matches
that do not wish to be a resource for that particular uri

IOW. a higher-priority uri matches (e.g. .h5), but the data is
actually PyTables, so HDFStore 'passes' on the processing. PyTables
is next in the chain so it will then process.

Note that a uri can be more fully-disambiguated via a head element
The default is ``hdfstore`` (this only matters for writing), reading
is unambiguous.

e.g.

```
....../file.h5 -> hdfstore
hdfstore://....../file.hdf5|h5 -> hdfstore
pytables://....../file.hdf5|h5 -> pytables
h5py://........../file.hdf5|h5 -> h5py
```


jreback:
@mrocklin @cpcloud 

ok, I updated this and finally fixed the segfault (was an import order issue).

I reverted a lot of the cleanup behavior, it will not raise NotImpementedError anymore as this is not necessary in the new impl.

Basically all of the hdf impl get wrapped in HDFFile/HDFTable. and that auto opens/closes for each operation. Its all dispatched thru create/append/discover/drop/cleanup. So I didn't actually have to change the impls much at all.

See new behavior at the top.

There are a million commits. It can all be squashed, but lmk.

mrocklin:
Hey Jeff, thanks for continuing to work on this.  Some feedback.

Biggish thoughts
---------------

This PR is fairly daunting and seems to invent a lot of concepts in order to support `HDFStore`.  I wonder if there is a simpler solution that we could approach first.

Creating new classes that wrap other libraries causes me to hesitate.  We've had a policy of playing nicely with other libraries rather than wrapping them and I think this has served us well so far.  In my mind `resource('foo.hdf5')` would yield an `h5py.File`, `tables.File`, or `pandas.HDFStore`, not a new into class.

If I were to approach this problem I would first try to solve it in a more isolated way that focused just on interacting with `HDFStore` in the way we've done previously with `h5py`.

Nitpicky things
-------------

Some simple things still fail.  Some examples:

```Python
In [28]: into('foo2.hdf5::/x', [1, 2, 3])
ValueError: need to pass a proper datashape to HDFStore

In [32]: x = into('foo2.hdf5::/x', [1, 2, 3], dshape=dshape('3 * int'))
ValueError: need to pass a proper datashape to HDFStore

# Make this file in master
In [4]: x = into('foo.hdf5::/x', [1, 2, 3])
In [5]: x
Out[5]: <HDF5 dataset "x": shape (3,), type "<i8">

# Back to development branch
In [33]: x = resource('foo.hdf5::/x')
In [34]: into(list, x)
KeyError: <class 'into.backends.hdf.HDFFile'>

In [40]: list(into(Iterator, uri_hdfstore))
ClosedNodeError: the node object is closed
```

I suspect that this mismatch is caused due to poor documentation and testing on my part and on trying to change too many things at once.

jreback:
>> Creating new classes that wrap other libraries causes me to hesitate. We've had a policy of playing nicely with other libraries rather than wrapping them and I think this has served us well so far. In my mind resource('foo.hdf5') would yield an h5py.File, tables.File, or pandas.HDFStore, not a new into class.

Well, if you *don't* do this we are back to square one. No way to properly clean up open file handles. This way everything is opened via a context manager and its nice and clean.

Further this hides the distinction between different hdf flavors. I would ask you why *wouldn't* you do this? You can easily customize the ``__str__`` if you want, provide nice accessors and a nice API. seems like the ideal solution to me. And if even dispatches cleanly on all of the registered methods.

>>In [28]: into('foo2.hdf5::/x', [1, 2, 3])
ValueError: need to pass a proper datashape to HDFStore

This is a single list, its not convertible to a dataframe.

>>In [32]: x = into('foo2.hdf5::/x', [1, 2, 3], dshape=dshape('3 * int'))
ValueError: need to pass a proper datashape to HDFStore
Same issue. I didn't expose Series (could but already big enough PR).

>>># Make this file in master
In [4]: x = into('foo.hdf5::/x', [1, 2, 3])
In [5]: x
Out[5]: <HDF5 dataset "x": shape (3,), type "<i8">

>>># Back to development branch
In [33]: x = resource('foo.hdf5::/x')
In [34]: into(list, x)
KeyError: <class 'into.backends.hdf.HDFFile'>

Same issue, these are simply not convertible to lists.

>>>In [43]: into('foo2.hdf5::/y', np.ones((5, 5)))
AttributeError: 'numpy.int64' object has no attribute 'startswith'

This again is simply not convertible to a frame. Yes its an ndarray, but hdfstore requires a DataFrame.

>>In [40]: list(into(Iterator, uri_hdfstore))
ClosedNodeError: the node object is closed

You cannot do this directly. An iterator requies an open file handle, so its done in a context manager. This is explicity handled by append.

mrocklin:
I think that handling file handles is probably out of scope.  I'd prefer to leave that responsibility to user.  If they give us file handles they should manage them.  The point you brought up earlier in the `into(foo, 'myfile.hdf5::/datapath')` is a good counter-example though.  We should clean up in this case.  I don't think we should worry about the other cases though.

Regarding the examples that no longer work do you not think that these use cases for `into` are important?  Into isn't only about pandas dataframes.  If HDFStore doesn't support them that's fine, but then it shouldn't be default behavior.

The `into(Iterator, ...)` is an important use case for Blaze.

jreback:
>>>I think that handling file handles is probably out of scope. I'd prefer to leave that responsibility to user. If they give us file handles they should manage them. The point you brought up earlier in the into(foo, 'myfile.hdf5::/datapath') is a good counter-example though. We should clean up in this case. I don't think we should worry about the other cases though.

You missed my point again. I said if a uri is given, then blaze should clean up. Your existing impl does not (and cannot not ever handle this properly). What I abstracted does.


>>>Regarding the examples that no longer work do you not think that these use cases for into are important? Into isn't only about pandas dataframes. If HDFStore doesn't support them that's fine, but then it shouldn't be default behavior.

HDFStore does not support non-pandas structures at all. I don't care if its the default. Though since pandas is pretty popular its prob used quite a bit.


>>>The into(Iterator, ...) is an important use case for Blaze.

You missed my point. This is supported, but you can only do it in context.

mrocklin:
I prefer not to wrap other classes because 

1.  It leads to (I think) a lot of code complexity
2.  It fractures the ecosystem and sets us apart 
3.  It limits inter-operation with other libraries and developers already familiar with those objects
4.  I think that it leads to greater maintenance costs

Blaze and into touch a lot of projects.  Wrapping all of them is a large task.  Playing nicely is, I think cleaner.

jreback:
Yet you wrap CSV. what's the difference?

I think closing resources is extremely important. You are losing site of the details which will make blaze successful.

mrocklin:
We didn't have a choice with CSV, there was no other project in the ecosystem to represent a csv object.  We do have a choice here.

I don't think that we're going to come to an agreement.  Moreover I think that this discussion has stopped being a productive use of time for either of us.  I'm going to close it.  I'll continue to add HDFStore support in #16 and raise issues on some of the other topics that you've brought up from your work here.  My apologies for allowing this to draw on.

PR #16, opened 2015-01-06
Support HDFStore - take two

mrocklin:
This is a work in progress.  It's an alternative implementation to #8 that tries to be just about supporting `HDFStore` in isolation.

Some issues came up

1.  HDFStore doesn't support empty datasets, this causes a bit of a hacky solution here.  It'd be nice if that was supported upstream
2.  I don't yet support groups, just flat HDF5 files
3.  PyTables doesn't like unicode and datashape usually assumes unicode by default.  This mismatch makes doing work rather finicky.  You have to specify kwarg stuff fairly often.  It might make sense to default to replace all instances of `string` with `string["A"]`.
4.  I have yet to really give it the run-around on lots of datasets
5.  Doesn't address any of the side concerns brought up in #8

On the plus side, it is short and snappy.  

Questions:

1. Is `HDFStore` intended to interact with datasets that don't fit the DataFrame model?  Should I bother to try to generalize or no?  
2.  How does `HDFStore` handle strings?  Only Python objects or is there some way to specify fixed-length?
3.  Is it possible to get particular chunks out of a `pd.io.pytables.FixedFrame` object?  I've been using the `HDFStore.select` method and none of the `chunksize`, `start/stop` kwargs work for this.

TODO (maybe future work):

- [ ] Raise informative errors if a file in `resource('.hdf5')` if a user tries to use a datetime datashape.  Suggest use of `hdfstore://` protocol
- [ ] Inspect metadata to see if we can identify HDFStore formatted hdf5 files automatically

### DIFF
--- a/into/convert.py
+++ b/into/convert.py
@@ -79,6 +79,12 @@ def numpy_chunks_to_numpy(c, **kwargs):
     return np.concatenate(list(c))
 
 
+@convert.register(chunks(np.ndarray), np.ndarray, cost=0.1)
+def numpy_to_chunks_numpy(x, chunksize=2**20, **kwargs):
+    return chunks(np.ndarray)(
+            lambda: (x[i:i+chunksize] for i in range(0, x.shape[0], chunksize)))
+
+
 def ishashable(x):
     try:
         hash(x)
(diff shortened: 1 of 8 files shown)


into/convert.py
```diff
@@ -79,6 +79,22 @@ def numpy_chunks_to_numpy(c, **kwargs):
...
 
+@convert.register(chunks(np.ndarray), np.ndarray, cost=0.5)
+def numpy_to_chunks_numpy(x, chunksize=2**20, **kwargs):
+    return chunks(np.ndarray)(
+            lambda: (x[i:i+chunksize] for i in range(0, x.shape[0], chunksize)))
+
+
+@convert.register(pd.DataFrame, chunks(pd.DataFrame), cost=1.0)
+def chunks_dataframe_to_dataframe(c, **kwargs):
+    return pd.concat(list(c), axis=0)
```
cpcloud: i would pass `**kwargs` in here and maybe put an `axis=0` default in the signature. there are other kwargs that are useful here: 

http://pandas.pydata.org/pandas-docs/stable/generated/pandas.tools.merge.concat.html

mrocklin: In this case I think we only want `axis=0`.  We're not replicating pd.concat generically (one should presumably just use `pd.concat` for that) and logically all of our chunks are row-wise part of the same table.  The other kwargs might be helpful though.

mrocklin:
Merging this for now for use by bokeh folks.  Sorry for the rush.

PR #18, opened 2015-01-08
Handle datetime64[> ns] and object series with record dshape

cpcloud:


cpcloud:
@mrocklin this fixes some things that I broke in blaze from #17

PR #28, opened 2015-01-15
rename chunks -> iterable

### LINKED ISSUE #23: Rename Chunks to IterableOf, add IndexableOf
The `Chunks` type is really just anything over which we can iterate.  It doesn't actually have to be used in a chunking workflow.  It might be best to rename it appropriately.

Separately, we might want to make an `IndexableOf` type that supports both iteration and also indexing.  Whereas `IterableOf(a)`takes an iterable or a function `f :: -> [a]`,  `IndexableOf(a)` might take either an indexable (like a list) or a function `f ::  int -> a` and a number of elements.  `IndexableOf` might also be extended to something n-dimensional, so possibly `f :: [int] -> a` and a shape rather than a number of elements.

mrocklin:
closes #23

### DIFF
--- a/into/__init__.py
+++ b/into/__init__.py
@@ -9,7 +9,7 @@ from .append import append
 from .resource import resource
 from .into import into
 from .drop import drop
-from .chunks import chunks, Chunks
+from .chunks import iterable, IterableOf
 from datashape import discover, dshape
 from collections import Iterator
 import numpy as np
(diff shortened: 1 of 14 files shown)


cpcloud:
@mrocklin should we get this in for the next release?

mrocklin:
No strong opinion.

cpcloud:
ok, i'll push to the next release. if i get everything else in then i'll get this in for 0.3.2

PR #35, opened 2015-01-18
AWS into

### LINKED ISSUE #34: AWS <-> AWS
From a private conversation:

Would be nice if `into` could move files in S3 to e.g., Redshift

```
data = into('redshift://conn_info::table', 's3://my/s3/bucket/*.csv')
```

cpcloud:
closes #34 

- [x] empty buckets
- [x] get schema from sqlalchemy
- [x] ~~multipart gzip~~
- [x] ~~single file gzip~~
- [x] ~~gzip with json and jsonlines~~
- [ ] tests for json
- [x] tests for jsonlines
- [x] ~~make a `Bucket` metaclass~~ not necessary ATM
- [x] handle temporary cleanup (handled by creating a temporary bucket and removing it each time, could be improved)
- [x] test SSH to S3

punting on gzip for now

see http://stackoverflow.com/questions/15754610/how-to-gzip-while-uploading-into-s3-using-boto and https://github.com/wrobstory/pgshift/blob/master/pgshift/pgshift.py#L143

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,11 +24,11 @@ install:
   - conda update conda
 
   # Install dependencies
-  - conda create -n test-environment python=$TRAVIS_PYTHON_VERSION pytest numpy pandas sqlalchemy h5py pip cython bcolz coverage networkx toolz multipledispatch pytables dynd-python pymongo paramiko
+  - conda create -n test-environment python=$TRAVIS_PYTHON_VERSION pytest numpy pandas sqlalchemy h5py pip cython bcolz coverage networkx toolz multipledispatch pytables dynd-python pymongo paramiko boto
   - source activate test-environment
 
   # Install various deps
-  - pip install pymongo pymysql psycopg2 paramiko sas7bdat boto
+  - pip install pymysql psycopg2 sas7bdat
 
   # Install DataShape
   - pip install --upgrade git+http://github.com/ContinuumIO/datashape
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,20 +27,13 @@ install:
   - conda create -n test-environment python=$TRAVIS_PYTHON_VERSION pytest numpy pandas sqlalchemy h5py pip cython bcolz coverage networkx toolz multipledispatch pytables dynd-python
   - source activate test-environment
 
-  # Install PyMongo
-  - pip install pymongo pymysql paramiko sas7bdat psycopg2
+  # Install various deps
+  - pip install pymongo pymysql psycopg2 paramiko sas7bdat boto
 
   # Install DataShape
   - pip install --upgrade git+http://github.com/ContinuumIO/datashape
 
-  # Install sas7bdat
-  - pip install sas7bdat
-
-  # install psycopg2 for postgres testing
-  - pip install psycopg2
-
-  # aws things
-  - pip install boto
+  # redshift sqlalchemy dialect
   - pip install --upgrade git+git://github.com/cpcloud/redshift_sqlalchemy
 
   # For bcolz
--- a/into/backends/aws.py
+++ b/into/backends/aws.py
@@ -26,7 +26,9 @@ sample = Dispatcher('sample')
 
 
 @memoize
-def get_s3_connection(aws_access_key_id, aws_secret_access_key, anon=False):
+def get_s3_connection(aws_access_key_id=None,
+                      aws_secret_access_key=None,
+                      anon=False):
     cfg = boto.Config()
 
     if aws_access_key_id is None:
--- a/into/backends/aws.py
+++ b/into/backends/aws.py
@@ -14,15 +14,12 @@ import boto
 import pandas as pd
 
 from into import discover, CSV, resource, append, convert, drop
-from ..utils import tmpfile, ext
+from ..utils import tmpfile, ext, sample
 
 try:
     from urlparse import urlparse
 except ImportError:
     from urllib.parse import urlparse
-
-from multipledispatch import Dispatcher
-sample = Dispatcher('sample')
 
 
 @memoize
--- a/into/backends/aws.py
+++ b/into/backends/aws.py
@@ -4,6 +4,7 @@ import os
 import uuid
 
 from contextlib import contextmanager
+from collections import Iterator
 
 try:
     from urlparse import urlparse
@@ -153,17 +154,22 @@ def discover_s3_line_delimited(c, length=8192, **kwargs):
         return discover(c.subtype(fn, **kwargs), **kwargs)
 
 
-@resource.register('s3://.*\.(csv|txt)', priority=18)
+@resource.register('s3://.*\.csv(\.gz)?', priority=18)
 def resource_s3_csv(uri, **kwargs):
     return S3(CSV)(uri, **kwargs)
 
 
-@resource.register('s3://.*\.json', priority=18)
+@resource.register('s3://.*\.txt(\.gz)?')
+def resource_s3_text(uri, **kwargs):
+    return S3(TextFile)(uri)
+
+
+@resource.register('s3://.*\.json(\.gz)?', priority=18)
 def resource_s3_json_lines(uri, **kwargs):
     return S3(JSONLines)(uri, **kwargs)
 
 
-@drop.register((S3(CSV), S3(JSON), S3(JSONLines)))
+@drop.register((S3(CSV), S3(JSON), S3(JSONLines), S3(TextFile)))
 def drop_s3(s3):
     s3.object.delete()
 
@@ -178,19 +184,28 @@ def drop_temp_s3(s3):
 @convert.register(Temp(CSV), (Temp(S3(CSV)), S3(CSV)))
 @convert.register(Temp(JSON), (Temp(S3(JSON)), S3(JSON)))
 @convert.register(Temp(JSONLines), (Temp(S3(JSONLines)), S3(JSONLines)))
-def s3_csv_to_temp_csv(s3, **kwargs):
+@convert.register(Temp(TextFile), (Temp(S3(TextFile)), S3(TextFile)))
+def s3_text_to_temp_text(s3, **kwargs):
     tmp_filename = '.%s.%s' % (uuid.uuid1(), ext(s3.path))
     s3.object.get_contents_to_filename(tmp_filename)
     return Temp(s3.subtype)(tmp_filename, **kwargs)
 
 
 @append.register(CSV, S3(CSV))
-def s3_csv_to_csv(csv, s3, **kwargs):
-    return append(csv, into(Temp(CSV), s3, **kwargs), **kwargs)
+@append.register(JSON, S3(JSON))
+@append.register(JSONLines, S3(JSONLines))
+@append.register(TextFile, S3(TextFile))
+def s3_text_to_text(data, s3, **kwargs):
+    return append(data, into(Temp(s3.subtype), s3, **kwargs), **kwargs)
 
 
 @append.register((S3(CSV), Temp(S3(CSV))), (S3(CSV), Temp(S3(CSV))))
-def s3_csv_to_s3_csv(a, b, **kwargs):
+@append.register((S3(JSON), Temp(S3(JSON))), (S3(JSON), Temp(S3(JSON))))
+@append.register((S3(JSONLines), Temp(S3(JSONLines))),
+                 (S3(JSONLines), Temp(S3(JSONLines))))
+@append.register((S3(TextFile), Temp(S3(TextFile))),
+                 (S3(TextFile), Temp(S3(TextFile))))
+def temp_s3_to_s3(a, b, **kwargs):
     a.object.bucket.copy_key(b.object.name, a.object.bucket.name,
                              b.object.name)
     return a
@@ -199,21 +214,24 @@ def s3_csv_to_s3_csv(a, b, **kwargs):
 @convert.register(Temp(S3(CSV)), (CSV, Temp(CSV)))
 @convert.register(Temp(S3(JSON)), (JSON, Temp(JSON)))
 @convert.register(Temp(S3(JSONLines)), (JSONLines, Temp(JSONLines)))
-def text_data_to_temp_s3_text_data(data, **kwargs):
+@convert.register(Temp(S3(TextFile)), (TextFile, Temp(TextFile)))
+def text_to_temp_s3_text(data, **kwargs):
     subtype = getattr(data, 'persistent_type', type(data))
     uri = 's3://%s/%s.%s' % (uuid.uuid1(), uuid.uuid1(), ext(data.path))
     return append(Temp(S3(subtype))(uri, **kwargs), data)
 
 
-@append.register(S3(CSV), (pd.DataFrame, chunks(pd.DataFrame), object))
-def anything_to_s3_csv(s3, o, **kwargs):
-    return into(s3, into(Temp(CSV), o, **kwargs), **kwargs)
+@append.register((S3(CSV), S3(JSON), S3(JSONLines)),
+                 (pd.DataFrame, chunks(pd.DataFrame), (list, Iterator)))
+def anything_to_s3_text(s3, o, **kwargs):
+    return into(s3, into(Temp(s3.subtype), o, **kwargs), **kwargs)
 
 
 @append.register(S3(JSONLines), (JSONLines, Temp(JSONLines)))
 @append.register(S3(JSON), (JSON, Temp(JSON)))
 @append.register(S3(CSV), (CSV, Temp(CSV)))
-def append_csv_to_s3(s3, data, **kwargs):
+@append.register(S3(TextFile), (TextFile, Temp(TextFile)))
+def append_text_to_s3(s3, data, **kwargs):
     s3.object.set_contents_from_filename(data.path)
     return s3
 
--- a/into/backends/aws.py
+++ b/into/backends/aws.py
@@ -15,8 +15,9 @@ from toolz import memoize, curry, pipe
 
 import boto
 
-from into import discover, CSV, resource, append, convert, drop, Temp, JSON, _Temp
+from into import discover, CSV, resource, append, convert, drop, Temp, JSON
 from into import JSONLines
+from into.backends.temp import _Temp
 from ..utils import tmpfile, ext, sample
 from ..utils import pmap, split
 
--- a/into/backends/aws.py
+++ b/into/backends/aws.py
@@ -122,22 +122,7 @@ def sample_s3_line_delimited(data, length=8192):
 
 @discover.register((S3(CSV), S3(JSONLines)))
 def discover_s3_line_delimited(c, length=8192, **kwargs):
-    """Discover CSV and JSONLines files from S3.
-
-    Examples
-    --------
-    >>> csv_dshape = discover(resource('s3://nyqpug/tips.csv'))
-    >>> print(csv_dshape)
-    var * {
-      total_bill: ?float64,
-      tip: ?float64,
-      sex: ?string,
-      smoker: ?string,
-      day: ?string,
-      time: ?string,
-      size: int64
-      }
-    """
+    """Discover CSV and JSONLines files from S3."""
     with sample(c, length=length) as fn:
         return discover(c.subtype(fn, **kwargs), **kwargs)
 
--- a/into/backends/aws.py
+++ b/into/backends/aws.py
@@ -2,7 +2,6 @@ from __future__ import print_function, division, absolute_import
 
 import os
 import uuid
-import atexit
 
 from contextlib import contextmanager
 
@@ -11,13 +10,15 @@ try:
 except ImportError:
     from urllib.parse import urlparse
 
+import pandas as pd
 from toolz import memoize, curry, pipe
+from toolz.curried import map
 
 import boto
 
 from into import discover, CSV, resource, append, convert, drop, Temp, JSON
-from into import JSONLines
-from into.backends.temp import _Temp
+from into import JSONLines, SSH, into
+
 from ..utils import tmpfile, ext, sample
 from ..utils import pmap, split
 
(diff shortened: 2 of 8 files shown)


cpcloud:
hm caching the `dshape` on network based non-database resources may be warranted here

into/backends/aws.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+@resource.register('s3://.*\.csv', priority=18)
+def resource_s3_csv(uri):
+    return s3(CSV)(uri)
+
+
+@convert.register(CSV, s3(CSV))
+def convert_s3_to_csv(bucket, **kwargs):
+    return CSV(bucket.data.path, **kwargs)
```
mrocklin: I think that ideally this downloads the file.  Perhaps that's the WIP part though.

into/backends/aws.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+@discover.register(s3(CSV))
+def discover_s3_csv(c, **kwargs):
+    return var * discover(c.data).subshape[0]
```
mrocklin: Ideally this downloads a tiny bit of the data, perhaps boto helps here.

into/backends/aws.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+@convert.register(pd.DataFrame, s3(CSV))
+def convert_csv_to_s3(csv):
+    return convert(pd.DataFrame, csv.data)
```
mrocklin: Generally having one convert function call another is an anti-pattern.  One should rely on the network for this.

into/backends/aws.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+@resource.register('s3://.*\.csv', priority=18)
+def resource_s3_csv(uri):
+    return S3(CSV)(uri)
+
+
+@convert.register(CSV, S3(CSV))
+def convert_s3_to_csv(csv, **kwargs):
+    """Download an S3 CSV to a local CSV file."""
```
mrocklin: Calling `convert` to create new persistent things (like file-backed resources) is currently something of a wart with `into`.  My tactic thus far has been to assume that user is doing something like 

    into('newfile.csv', 's3://.../oldfile.csv')

And so we'll actually be calling `append` on to an empty and just-created `CSV` object.

I'm not sure that this is the best way forward though, thoughts?

into/backends/aws.py
```diff
@@ -0,0 +1,184 @@
...
+                           anon=anon)
+
+
+@convert.register(S3(CSV), CSV)
+def convert_csv_to_s3(csv,
+                      bucket_name=None,
+                      aws_access_key_id=None,
+                      aws_secret_access_key=None,
+                      **kwargs):
+    assert bucket_name is not None, 'must provide a bucket name'
```
mrocklin: Is it possible to create empty S3 buckets and then append on to them?  If not, what about making some sort of token for a just-created S3 bucket?

into/backends/aws.py
```diff
@@ -0,0 +1,184 @@
...
+    # TODO: is Content-Type useful here?
+    resp = conn.make_request('GET',
+                             bucket=bucket,
+                             key=key,
+                             headers={'Range': 'bytes=%d-%d' % (start, stop)})
+    raw = resp.read()
+    raw = raw[:raw.rindex(b'\n')]
+    with tempfile.NamedTemporaryFile(mode='wb', delete=False) as f:
+        f.write(raw)
+    return CSV(f.name, **kwargs)
```
mrocklin: Should maybe be a `contextmanager` to remove the temporary file?  My HDFS solution to this is here https://github.com/ContinuumIO/into/pull/44/files#diff-336aefd0b6a5fb97d9a47767a30175d8R56

into/backends/aws.py
mrocklin: To be more clear.  I actually expect the LHS to turn into `resource('newfile.csv') -> CSV(...)`.  So we don't expect a call to `convert`, we expect a call to `append`.

into/backends/aws.py
cpcloud: You can create empty s3 buckets, but it doesn't make sense to create unnamed ones

into/backends/aws.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+@append.register(CSV, S3(CSV))
+def append_s3_to_csv(csv, s3csv, **kwargs):
+    """Download an S3 CSV to a local CSV file."""
+    s3csv.object.get_contents_to_filename(csv.path)
+    return append(csv, CSV(csv.path, **kwargs))
+
+
+@sample.register(S3(CSV))
```
mrocklin: Does this work for all `_S3`?

into/backends/aws.py
mrocklin: Does the `append` trick work here as well if `resource('s3://...')` makes an empty bucket?

into/backends/aws.py
```diff
@@ -0,0 +1,174 @@
...
+    s, substring : str
+
+    Returns
+    -------
+    stripped : str
+        `s` without `substring`
+
+    Examples
+    --------
+    >>> uri_to_path('s3://nyqpug/tips.csv')  # doctest: +SKIP
```
mrocklin: Why the doctest skip?

into/backends/aws.py
```diff
@@ -0,0 +1,174 @@
...
+    'nyqpug/tips.csv'
+    >>> uri_to_path('http://nyqpug/tips.csv')  # doctest: +SKIP
+    'nyqpug/tips.csv'
+    """
```
mrocklin: Is `uri.split('://')[1]` a sufficient implementation?

into/backends/sql_csv.py
```diff
@@ -83,6 +83,38 @@ def copy_mysql(dialect, tbl, csv):
...
+    import boto
+    from into.backends.aws import S3
+    from redshift_sqlalchemy.dialect import CopyCommand
+except ImportError:
+    pass
+else:
+    # TODO: kwargs passing? there are some dialect specific things that might
+    # be useful here
+    @copy_command.register('redshift.*')
+    def copy_redshift(dialect, tbl, csv, schema_name='public'):
```
mrocklin: Can the schema_name be grabbed somehow from the `tbl`?  Perhaps from the associated engine (`tbl.bind`)?

into/backends/aws.py
cpcloud: got rid of this function

into/backends/sql_csv.py
cpcloud: possibly, i'll look into it

into/backends/aws.py
cpcloud: will look into it

into/backends/aws.py
cpcloud: not necessarily, because i search for a line delimiter.

for example minified JSON would fail and it's fairly common to minify it

cpcloud:
@mrocklin there's a slight issue coming up with temporary files. In addition to creating a temporary key I must also create a temporary bucket. One solution is to make a global temporary bucket for the current Python process and have an `atexit` handler which tries to delete it on successful termination of a Python program.

mrocklin:
My guess is that a temporary key is like a temporary filename and so perhaps `uuid` would be good here.

`Temp` is supposed to be able to handle the temporary bucket problem for you.  The garbage collector serves as the cleanup service.  Upon clean exit of the python process all refs eventually get cleaned up.  Any object that inherits from `Temp` will call `drop` on itself at this time.  

1.  Implement `drop(S3(CSV))` to remove the temporary bucket (I think you already have this)
2.  Create a `Temp(S3(CSV))` then exit the python process and see if it persists.

Not sure if this answers your problem or not.  I also might not fully understand the situation with S3 bucket.

cpcloud:
So the deletion of a temporary bucket is fine. That wasn't really the issue, my mistake. The issue arises when we want to drop a resource from S3 *at all*. What if I have 10 files in a bucket and I only want to remote a single file? I don't want to clobber the entire bucket when this happens.

One solution is to have two `drop` implementations. An S3 one, which deletes only the key and another Temp(S3) that deletes the entire bucket.

mrocklin:
My thought was that a single `S3(CSV)` corresponds to a key in a bucket.  `drop(my_s3_csv)` deletes the key from that bucket.

So, if I had a `Temp(S3(CSV))` that was given an explicit bucket and key then it would delete that key from that bucket when it was cleaned up.  This should happen with code as is.  

Perhaps the confusion is where to put temporary s3 things when we need to invent names for them.  It seems like we also need to create a bucket and manage that.  I don't have any strong ideas on how to handle this.

We could create a `Temp(Bucket)` (perhaps treat `Bucket` like  `S3(Directory)`?) and then give a reference of it to each of the `S3(CSV)`s that we put in to that bucket.  Then, once those references disappear then the Bucket itself would disappear.

into/backends/aws.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+@resource.register('s3://.*\.(csv|txt)', priority=18)
+def resource_s3_csv(uri, **kwargs):
+    return S3(CSV)(uri, **kwargs)
+
+
+@resource.register('s3://.*\.json', priority=18)
+def resource_s3_csv(uri, **kwargs):
+    return S3(JSON)(uri, **kwargs)
```
mrocklin: Should this be `JSONLines` by default?

into/backends/aws.py
```diff
@@ -0,0 +1,206 @@
...
+        yield fn
+
+
+@discover.register(S3(CSV))
+def discover_s3_csv(c, length=8192, **kwargs):
+    with sample(c, length=length) as fn:
+        return discover(CSV(fn, **kwargs), **kwargs)
+
+
+@resource.register('s3://.*\.(csv|txt)', priority=18)
```
mrocklin: `(\.gz)?`  ?

into/backends/aws.py
```diff
@@ -0,0 +1,206 @@
...
+def resource_s3_csv(uri, **kwargs):
+    return S3(CSV)(uri, **kwargs)
+
+
+@resource.register('s3://.*\.json', priority=18)
+def resource_s3_csv(uri, **kwargs):
+    return S3(JSON)(uri, **kwargs)
+
+
+@drop.register((S3(CSV), S3(JSON), S3(JSONLines)))
```
mrocklin: I added `TextFile` in #56 and started including it in these lists.

into/backends/aws.py
```diff
@@ -0,0 +1,206 @@
...
+@append.register(CSV, S3(CSV))
+def s3_csv_to_csv(csv, s3, **kwargs):
+    return append(csv, into(Temp(CSV), s3, **kwargs), **kwargs)
+
+
+@append.register((S3(CSV), Temp(S3(CSV))), (S3(CSV), Temp(S3(CSV))))
+def s3_csv_to_s3_csv(a, b, **kwargs):
+    a.object.bucket.copy_key(b.object.name, a.object.bucket.name,
+                             b.object.name)
+    return a
```
mrocklin: This and the function above could maybe be extended to the other text types.

into/backends/aws.py
```diff
@@ -0,0 +1,206 @@
...
+
+@append.register(S3(CSV), (pd.DataFrame, chunks(pd.DataFrame), object))
+def anything_to_s3_csv(s3, o, **kwargs):
+    return into(s3, into(Temp(CSV), o, **kwargs), **kwargs)
+
+
+@append.register(S3(JSONLines), (JSONLines, Temp(JSONLines)))
+@append.register(S3(JSON), (JSON, Temp(JSON)))
+@append.register(S3(CSV), (CSV, Temp(CSV)))
+def append_csv_to_s3(s3, data, **kwargs):
```
mrocklin: probably grep for `csv` and change to `file` in a few places.

into/backends/csv.py
```diff
@@ -257,6 +250,16 @@ def convert_dataframes_to_temporary_csv(data, **kwargs):
     return append(csv, data, **kwargs)
 
 
+@append.register(CSV, Temp(CSV))
+def temp_csv_to_csv(a, b, **kwargs):
+    return append(a, CSV(b.path), **kwargs)
```
mrocklin: Under certain circumstances this could just transfer text.

mrocklin:
```
with open(b.path) as f:
    with open(a.path, 'a') as f:
        ...
```

mrocklin: In general we want to avoid parsing/emiting csv when possible.

into/backends/csv.py
```diff
@@ -257,6 +250,16 @@ def convert_dataframes_to_temporary_csv(data, **kwargs):
...
+
+
+@convert.register(Temp(CSV), CSV)
+def csv_to_temp_csv(csv, **kwargs):
+    return Temp(CSV)(csv.path, **kwargs)
```
mrocklin: Beware, this might cause the original file to be lost.  Probably we want a new filename and an os-level copy.

mrocklin: Or one could ask the question why we need to convert a `CSV` to a `Temp(CSV)`.  For what operations do we need this?

into/utils.py
```diff
@@ -206,3 +211,82 @@ def ignoring(*exceptions):
...
+@curry
+def pmap(f, iterable):
+    """Map `f` over `iterable` in parallel using a ``ThreadPool``.
+    """
+    p = ThreadPool()
+    try:
+        result = p.map(f, iterable)
+    finally:
+        p.terminate()
+    return result
```
mrocklin: Slick

into/utils.py
```diff
@@ -206,3 +211,82 @@ def ignoring(*exceptions):
...
+    """
+    i, filename, data = triple
+    with writer(filename, mode='wb') as f:
+        f.write(data)
+    return i, filename
+
+
+def gentemp(it, suffix=None, start=0):
+    """Generate an index, a temporary file, and return data for each element
+    in `it
```
mrocklin: I think that docstring rendering systems will like you if the title line is a single line.  Might be able to squeeze this in if you drop the title to the line below the `"""`

into/utils.py
```diff
@@ -206,3 +211,82 @@ def ignoring(*exceptions):
...
+
+    Parameters
+    ----------
+    it : iterable
+    suffix : str
+        Suffix to add to each temporary file's name
+    """
+    for i, data in enumerate(it, start=start):  # aws needs parts to start at 1
+        _, fn = tempfile.mkstemp(suffix=suffix)
```
mrocklin: `_` may be an open file handle.  Windows may not like it if other operations try to `open(fn)`

into/backends/aws.py
cpcloud: I'm not sure. How would we get traditional JSON files as S3 resources?

mrocklin: We would have to use the types explicitly.

cpcloud:
@mrocklin this could use your review, sorry about the huge number of commits. only thing i'm not testing right now is regular (non-newline-delimited) JSON

.travis.yml
```diff
@@ -27,12 +27,15 @@ install:
   - conda create -n test-environment python=$TRAVIS_PYTHON_VERSION pytest numpy pandas sqlalchemy h5py pip cython bcolz coverage networkx toolz multipledispatch pytables dynd-python pymongo paramiko
   - source activate test-environment
 
-  # Install PyMongo
-  - pip install sas7bdat psycopg2 pymysql
+  # Install various deps
+  - pip install pymongo pymysql psycopg2 paramiko sas7bdat boto
```
mrocklin: I've moved pymongo and paramiko to conda

.travis.yml
```diff
@@ -27,12 +27,15 @@ install:
...
 
   # Install DataShape
   - pip install --upgrade git+http://github.com/ContinuumIO/datashape
 
+  # redshift sqlalchemy dialect
+  - pip install --upgrade git+git://github.com/cpcloud/redshift_sqlalchemy
```
mrocklin: What is the long-term plan for this?

into/backends/aws.py
```diff
@@ -0,0 +1,248 @@
...
+
+    if aws_secret_access_key is None:
+        aws_secret_access_key = os.environ.get('AWS_SECRET_ACCESS_KEY')
+
+    # anon is False but we didn't provide any credentials so try anonymously
+    anon = (not anon and
+            aws_access_key_id is None and
+            aws_secret_access_key is None)
+    return boto.connect_s3(aws_access_key_id, aws_secret_access_key,
+                           anon=anon)
```
mrocklin: Can this go stale?  I ended up having to do the caching by hand in `ssh`

into/backends/aws.py
```diff
@@ -0,0 +1,248 @@
...
+def S3(cls):
+    """Parametrized S3 bucket Class
+
+    Examples
+    --------
+    >>> import sqlalchemy
+    >>> S3(CSV)
+    S3(CSV)
+    >>> S3(sqlalchemy.Table)
+    """
```
mrocklin: Recommend putting the docstring on the class in case people ask for help on the object.  I usually then copy over `__doc__`

into/backends/aws.py
```diff
@@ -0,0 +1,248 @@
...
+            aws_access_key_id is None and
+            aws_secret_access_key is None)
+    return boto.connect_s3(aws_access_key_id, aws_secret_access_key,
+                           anon=anon)
+
+
+class _S3(object):
+    def __init__(self, uri, s3=None, aws_access_key_id=None,
+                 aws_secret_access_key=None, *args, **kwargs):
+        result = urlparse(uri)
```
mrocklin: At some point we should probably create really informative errors around the use of URIs.  (not that my solutions are good at this either)

into/backends/aws.py
```diff
@@ -0,0 +1,248 @@
...
+@memoize
+def S3(cls):
+    """Parametrized S3 bucket Class
+
+    Examples
+    --------
+    >>> import sqlalchemy
+    >>> S3(CSV)
+    S3(CSV)
+    >>> S3(sqlalchemy.Table)
```
mrocklin: Can one put a `sqlalchemy.Table` on S3?  Perhaps also show how to create an `S3(CSV)` object explicitly

    >>> S3(CSV)(what-uri-is-appropriate?,  what kwargs should I know about?)

into/backends/aws.py
```diff
@@ -0,0 +1,248 @@
...
+    """
+    return type('S3(%s)' % cls.__name__, (_S3, cls), {'subtype': cls})
+
+
+@sample.register((S3(CSV), S3(JSONLines)))
+@contextmanager
+def sample_s3_line_delimited(data, length=8192):
+    """Get bytes from `start` to `stop` of `key` from the S3 bucket `bucket`.
```
mrocklin: This title seems to be out-of-sync with the options.

into/backends/aws.py
```diff
@@ -0,0 +1,248 @@
...
+
+    """
+    headers = {'Range': 'bytes=0-%d' % length}
+    raw = data.object.get_contents_as_string(headers=headers)
+
+    # this is generally cheap as we usually have a tiny amount of data
+    try:
+        index = raw.rindex(b'\r\n')
+    except ValueError:
+        index = raw.rindex(b'\n')
```
mrocklin: neat trick

into/backends/aws.py
```diff
@@ -0,0 +1,248 @@
...
+
+
+@discover.register((S3(CSV), S3(JSONLines)))
+def discover_s3_line_delimited(c, length=8192, **kwargs):
+    """Discover CSV and JSONLines files from S3.
+
+    Examples
+    --------
+    >>> csv_dshape = discover(resource('s3://nyqpug/tips.csv'))
+    >>> print(csv_dshape)
```
mrocklin: Will this fail doctests on a machine without the same boto file?  Perhaps skip?

into/backends/aws.py
```diff
@@ -0,0 +1,248 @@
...
+      time: ?string,
+      size: int64
+      }
+    >>> json_dshape = discover(resource('s3://nyqpug/tips.json'))
+    >>> names = list(map(str, sorted(json_dshape.measure.names)))
+    >>> names
+    ['day', 'sex', 'size', 'smoker', 'time', 'tip', 'total_bill']
+    >>> types = [json_dshape.measure[name] for name in names]
+    >>> from pprint import pprint
+    >>> pprint(types)
```
mrocklin: Are these doctests or tests?

into/backends/aws.py
cpcloud: Nope, this is a public read-only S3 bucket that I manage. This should only fail if I get hit by a :bus:

into/backends/aws.py
```diff
@@ -0,0 +1,248 @@
...
+
+
+@resource.register('s3://.*\.txt(\.gz)?', priority=18)
+def resource_s3_text(uri, **kwargs):
+    return S3(TextFile)(uri)
+
+
+@resource.register('s3://.*\.json(\.gz)?', priority=18)
+def resource_s3_json_lines(uri, **kwargs):
+    return S3(JSONLines)(uri, **kwargs)
```
mrocklin: Thanks for extending the functions below to `TextFile`.  The ones above could use some love too.

If you can think of a way to cleanly avoid code duplication without too much extra complexity I'd be thrilled.  This has me stumped.

into/backends/aws.py
cpcloud: Whoa, I didn't even see that I'd done this.

into/backends/aws.py
cpcloud: The JSON version should be a test

into/backends/aws.py
```diff
@@ -0,0 +1,248 @@
...
+    return into(s3, into(Temp(s3.subtype), o, **kwargs), **kwargs)
+
+
+@append.register(S3(JSONLines), (JSONLines, Temp(JSONLines)))
+@append.register(S3(JSON), (JSON, Temp(JSON)))
+@append.register(S3(CSV), (CSV, Temp(CSV)))
+@append.register(S3(TextFile), (TextFile, Temp(TextFile)))
+def append_text_to_s3(s3, data, **kwargs):
+    s3.object.set_contents_from_filename(data.path)
+    return s3
```
mrocklin: Perhaps many of the lines above could be handled with partial evaluation.

I started doing this in `ssh` but didn't really get in to it whole-hog.  https://github.com/ContinuumIO/into/blob/master/into/backends/ssh.py#L204-L215  it might be a way to avoid the giant lists of decorators.  For now though I say we wait until the problem gets worse.

cpcloud:
@mrocklin One thing to note here is that I left my file splitting code around in `utils`. I didn't want to totally kill it, but it's not being used for anything right now. I can gut it and move it to a gist or something.

cpcloud:
This is what `gentmp`, `write`, `split` and `pmap` are for

into/backends/aws.py
```diff
@@ -0,0 +1,248 @@
...
+@append.register(S3(JSON), (SSH(JSON), HDFS(JSON)))
+@append.register(S3(JSONLines), (SSH(JSONLines), HDFS(JSONLines)))
+@append.register(S3(CSV), (SSH(CSV), HDFS(CSV)))
+@append.register(S3(TextFile), (SSH(TextFile), HDFS(TextFile)))
+@append.register(HDFS(JSON), S3(JSON))
+@append.register(HDFS(JSONLines), S3(JSONLines))
+@append.register(HDFS(CSV), S3(CSV))
+@append.register(HDFS(TextFile), S3(TextFile))
+def remote_text_to_s3_text(a, b, **kwargs):
+    return into(a, into(Temp(b.subtype), b, **kwargs), **kwargs)
```
mrocklin: Historically I've been using  `convert` and `append` directly in internal code.  Not sure there is any reason though.

into/backends/sql_csv.py
```diff
@@ -1,19 +1,25 @@
...
 from ..regex import RegexDispatcher
 from ..append import append
 from .csv import CSV
-from multipledispatch import MDNotImplementedError
-import os
-import datashape
-import sqlalchemy
-import subprocess
+from ..temp import Temp
+from ..into import into
```
mrocklin: As above, I've been avoiding bringing `into` into internal code.  This breaks that pattern.  I'm not sure if it's a pattern worth keeping.

into/backends/tests/test_aws.py
```diff
@@ -0,0 +1,253 @@
...
+from into import into, resource, S3, discover, CSV, drop, append, SSH
+from into.backends.aws import get_s3_connection
+from into.utils import tmpfile, filetext
+import pandas as pd
+import pandas.util.testing as tm
+import datashape
+
+sa = pytest.importorskip('sqlalchemy')
+boto = pytest.importorskip('boto')
+pytest.importorskip('psycopg2')
```
mrocklin: Why do we need this driver in particular?

into/backends/tests/test_aws.py
```diff
@@ -0,0 +1,253 @@
...
+
+@pytest.fixture(scope='module')
+def conn():
+    # requires that you have a config file or envars defined for credentials
+    # this code makes me hate exceptions
+    try:
+        conn = get_s3_connection()
+    except S3ResponseError:
+        pytest.skip('unable to connect to s3')
+    else:
```
mrocklin: Can we denest the `else`'s?

mrocklin: Oh, or will Python keep chugging away on this function even after the pytest.skip

into/backends/tests/test_aws.py
```diff
@@ -0,0 +1,253 @@
...
+        drop(resource(b))
+
+
+@pytest.yield_fixture
+def s3_json_bucket(conn):
+    b = 's3://%s/%s.json' % (test_bucket_name, next(_tmps))
+    try:
+        yield b
+    finally:
+        drop(resource(b))
```
mrocklin: Can these be one fixture with arguments like `'csv'`, `'txt'`, ...?

mrocklin: With context managers I would do something like the following:

```Python
@contextmanger
def s3_bucket(extension):
    with conn():
        b = 's3://%s/%s.%s' % (test_bucket_name, next(_tmps), extension)
        try:
            yield b
        finally:
            drop(reource(b))
```

cpcloud: Yep I think a context manager is cleaner here

into/backends/tests/test_aws.py
```diff
@@ -0,0 +1,253 @@
...
+        df.to_csv(fn, index=False)
+        s3 = into(s3_bucket, CSV(fn))
+    result = into(pd.DataFrame, s3)
+    tm.assert_frame_equal(df, result)
+
+
+def test_s3_to_redshift(db):
+    redshift_uri = '%s::tips' % db
+    s3 = resource(tips_uri)
+    table = into(redshift_uri, s3, dshape=discover(s3))
```
mrocklin: Why did we need the `dshape=` kwarg here?

into/backends/tests/test_aws.py
```diff
@@ -0,0 +1,253 @@
...
+    ds = datashape.dshape("""
+    var * {
+        total_bill: ?float64,
+        tip: ?float64,
+        sex: ?string,
+        smoker: ?string,
+        day: ?string,
+        time: ?string,
+        size: int64
+    }""")
```
mrocklin: maybe `assert discover(table) == discover(s3)` to avoid having to tweak this if we make changes to exactly what the datashapes should be?  Makes the test a little more *unit*y.

into/backends/tests/test_aws.py
```diff
@@ -0,0 +1,253 @@
...
+
+def test_frame_to_s3_to_frame(s3_bucket):
+    s3_csv = into(s3_bucket, df)
+    result = into(pd.DataFrame, s3_csv)
+    tm.assert_frame_equal(result, df)
+
+
+def test_csv_to_redshift(tmpcsv, temp_tb):
+    table = into(temp_tb, tmpcsv)
+    assert table.count().execute().scalar() == 3
```
mrocklin: `== len(into(list, tmpcsv))`?

into/backends/aws.py
mrocklin: Well I'll be damned.  Ain't that somethin'

Just works right outta the box.

mrocklin:
> @mrocklin One thing to note here is that I left my file splitting code around in utils. I didn't want to totally kill it, but it's not being used for anything right now. I can gut it and move it to a gist or something.

Yeah, I noticed.  They seem simple, self contained, potentially useful, and out of the way.  I say we leave them in for a few months and see if they're useful.  We'll remove next time we cull dead code (if they're still dead).

into/backends/tests/test_aws.py
mrocklin: Wasn't necessarily suggesting a context manager, more suggesting some use of configuring these some how.  What's the typical way to do this with fixtures?

mrocklin:
In general this looks really good.  Definitely cleaner than my hack at HDFS.  

We should host example data on S3 and just have users download it in tutorials with `into`.

mrocklin:
@quasiben was interested in loading a Hive table from S3.  I wonder if he can point us to a good resource or example.

into/backends/aws.py
mrocklin: This feels like an antipattern.  usually we don't explicitly encode paths in the network.  Shouldn't the following implementations have this covered?

    convert  TMP(o) <- SSH(o)
    append: S3(o) <- o

into/backends/tests/test_aws.py
cpcloud: wasn't necessary

into/backends/tests/test_aws.py
cpcloud: yep

into/backends/tests/test_aws.py
cpcloud: agreed

into/backends/tests/test_aws.py
cpcloud: couldn't find a way to do this with fixtures, i don't think you can call them like functions

into/backends/tests/test_aws.py
cpcloud: not sure what will happen there, i'll look into it

into/backends/tests/test_aws.py
cpcloud: this is specified in the `setup.py` for `redshift_sqlalchemy`
https://github.com/cpcloud/redshift_sqlalchemy/blob/master/setup.py#L21

into/backends/aws.py
cpcloud: it probably is, i did this because i was getting ambiguity warnings. there's likely some redundancy on my part that i need to clean up

mrocklin: Ah, I see, I ran into those too. 

For `SSH <- S3` I wonder, can we assume that any machine into which we're ssh-ed also has `wget`?

I suggest the following fall back solution for ambiguities

    append.register(HDFS(CSV), S3(CSV))(append.dispatch(HDFS(CSV), object))

This says, treat `HDFS(CSV) <- S3(CSV)` as you would treat `HDFS(CSV) <- object`.  Probably exactly what you wrote down will happen, but the graph will be in control.

.travis.yml
mrocklin: Perhaps issue a pull request and ask him to update pypi?

cpcloud: created a PR here: https://github.com/binarydud/redshift_sqlalchemy/pull/27

into/backends/aws.py
cpcloud: What do you mean by stale in this context? I'm assuming that you won't be accessing things with different credentials on the same machine (note that you *can* do this, I deemed more-than-minimal authentication schemes as low priority while I was working on this). With SSH it matters more because the information to connect is specified in the URI whereas here it isn't.

into/backends/aws.py
cpcloud: done

into/backends/aws.py
cpcloud: fixed

into/backends/aws.py
cpcloud: fixed, moved to its own test

into/backends/aws.py
mrocklin: With ssh we might need to reconnect if the connection drops.  It's a single long-lived connection.  Presumably boto doesn't work like this though.

cpcloud: Ah, I see. I'm not sure if it works like that or not. I basically have a single long-lived connection as well. I'll look into this tomorrow in more detail.

into/backends/aws.py
mrocklin: Oh, actually, these should probably just `raise MDNotImplementedError()`.  Then multipledispatch will find the next best thing.

into/utils.py
```diff
@@ -11,6 +12,11 @@
 from .compatibility import unicode
 
 
+def ext(filename):
+    _, e = os.path.splitext(filename)
+    return e.lstrip(os.extsep)
```
mrocklin: Does this behave for `myfile.csv.gz`?

into/backends/aws.py
mrocklin: Of the comments still remaining I think that this is the only one that I really care about.

cpcloud: cool, on it now

into/utils.py
cpcloud: It returns `gz`. Would you want it to return `csv.gz`?

cpcloud: We use `ext(filename) == 'gz'` in a few places, so changing that behavior would break some things.

cpcloud: Though we could change that code to be `ext(filename).endswith('gz')`

mrocklin:
Generally +1 other than the ssh thing.

mrocklin:
Then I'm +1 with the caveat that we check binstar-build tests after merge

cpcloud:
cool

cpcloud:
running on my vm as well

cpcloud:
ran on my vm and can't get a call to `get_bucket()` to succeed in a reasonable amount of time, so skipping on windows

mrocklin:
Sounds good.  That one I suspect will be desired by people though.  Raise an issue for future work?

cpcloud:
bombs away

PR #37, opened 2015-01-19
Append to a table from a select statement

### LINKED ISSUE #36: Append sqlalchemy.Select to sqlalchemy.Table
Sometimes we might want to do a computation and then keep the result in the database as another table.  To do this Blaze needs an `append(sqlalchemy.Table, sqlalchemy.Select)` implementation.  Presumably there is some way to generate an approprate `INSERT INTO ...` statement through sqlalchemy.

cpcloud:
closes #36

### DIFF
--- a/into/backends/sql.py
+++ b/into/backends/sql.py
@@ -254,6 +254,17 @@ def append_anything_to_sql_Table(t, o, **kwargs):
     return append(t, convert(Iterator, o, **kwargs), **kwargs)
 
 
+@append.register(sa.Table, sa.sql.Select)
+def append_select_statement_to_sql_Table(t, o, **kwargs):
+    assert o.bind.has_table(t.name), 'tables must come from the same database'
+
+    query = t.insert().from_select(o.columns.keys(), o)
+
+    with o.bind.connect() as conn:
+        conn.execute(query)
+    return t
+
+
 @resource.register('(.*sql.*|oracle)(\+\w*)?://.+')
 def resource_sql(uri, *args, **kwargs):
     kwargs2 = keyfilter(keywords(sa.create_engine).__contains__,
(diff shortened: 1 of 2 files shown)


into/backends/tests/test_sql.py
```diff
@@ -185,3 +188,30 @@ def test_resource_on_dialects():
...
+                    (314.14, 'Hope'),
+                    (235.43, 'Bob')], dtype=[('amount', 'float64'),
+                                             ('name', 'S5')])
+    raw2 = np.array([(800.0, 'Joe'),
+                     (914.14, 'Alice'),
+                     (1235.43, 'Ratso')], dtype=[('amount', 'float64'),
+                                                 ('name', 'S5')])
+    t = into('%s::t' % sqlite_file, raw)
+    s = into('%s::s' % sqlite_file, raw2)
+    t = append(t, s.select())
```
mrocklin: Hrm, now that I think about this I bet we could make this work in either case by dispatching on `sqlalchemy.Selectable` rather than `sqlalchemy.Select`.

into/backends/sql.py
```diff
@@ -254,6 +254,17 @@ def append_anything_to_sql_Table(t, o, **kwargs):
...
 
+@append.register(sa.Table, sa.sql.Select)
+def append_select_statement_to_sql_Table(t, o, **kwargs):
+    assert o.bind.has_table(t.name), 'tables must come from the same database'
+
+    query = t.insert().from_select(o.columns.keys(), o)
+
+    with o.bind.connect() as conn:
+        conn.execute(query)
+    return t
```
mrocklin: Pretty slick.

mrocklin:
Merging soon if no comment.

PR #41, opened 2015-01-20
SAS7BDAT backend

talumbau:
Following [this](https://github.com/pydata/pandas/issues/4052) conversation, it appears that [sas7bdat](http://git.pyhacker.com/sas7bdat) now supports Python 3. This leverages the latest package, which can read compressed and uncompressed data. Some notes:

Since sas7bdat doesn’t write files, I’m not sure how to do the test data, other than put it in the repo, which seems odd

I don’t have a sense about cost weights. The conversions are unlikely to be a “middle step” though so it probably doesn’t matter.

This can get better if I get access to data that is more than either ‘string’ or ‘numeric’ types

Renaming `sql_csv.py:excute_copy` -> `execute_copy` causes an error, so I changed it to a (hopefully) meaningful name, just not the same name as the decorator

### DIFF
--- a/into/backends/sas.py
+++ b/into/backends/sas.py
@@ -1,16 +1,13 @@
 from __future__ import absolute_import, division, print_function
 
-try:
-    import sas7bdat
-    from sas7bdat import SAS7BDAT
-except ImportError:
-    sas7bdat = None
-    SAS7BDAT = type(None)
+import sas7bdat
+from sas7bdat import SAS7BDAT
 
 import datashape
-from datashape import discover, dshape, var
+from datashape import discover, dshape, var, Record, date_, datetime_
 from collections import Iterator
 import pandas as pd
+from .pandas import coerce_datetimes
 from ..append import append
 from ..convert import convert
 from ..resource import resource
@@ -23,21 +20,26 @@ def resource_sas(uri, **kwargs):
 
 @discover.register(SAS7BDAT)
 def discover_sas(f, **kwargs):
-    f.skip_header = True
-    ln = next(f.readlines())
-    ds = discover(ln)
-    cols = [col.name.decode("utf-8") for col in f.header.parent.columns]
-    types = [_type.strip() for _type in str(ds).strip("()").split(',')]
-    measure = ",".join(" " + col + ":" + _type for col, _type in zip(cols, types))
-    return var * dshape("{" + measure + "}")
+    lines = f.readlines()
+    next(lines) # burn header
+    ln = next(lines)
+    types = map(discover, ln)
+    names = [col.name.decode("utf-8") for col in f.header.parent.columns]
+    return var * Record(list(zip(names, types)))
 
 
 @convert.register(pd.DataFrame, SAS7BDAT, cost=4.0)
 def sas_to_DataFrame(s, dshape=None, **kwargs):
-    return s.to_data_frame()
+    df = s.to_data_frame()
+    if any(typ in (date_, datetime_) for typ in dshape.measure.types):
+        df = coerce_datetimes(df)
+    df = df[s.column_names]  # Reorder names to match sasfile
+    return df
 
 
 @convert.register(Iterator, SAS7BDAT, cost=1.0)
 def sas_to_iterator(s, **kwargs):
-    s.skip_header = True
-    return s.readlines()
+    lines = s.readlines()
+    if not s.skip_header:
+        next(lines)  # burn
+    return lines
--- a/into/backends/tests/test_sas.py
+++ b/into/backends/tests/test_sas.py
@@ -18,49 +18,50 @@ from numpy import dtype
 
 cur_path = os.path.abspath(os.path.dirname(__file__))
 test_path = os.path.join(cur_path, 'airline.sas7bdat')
-
-
-@pytest.yield_fixture
-def sasfile():
-    with SAS7BDAT(test_path) as f:
-        yield f
+sasfile = SAS7BDAT(test_path)
 
 
 columns = ("DATE", "AIR", "mon1", "mon2", "mon3", "mon4", "mon5", "mon6",
            "mon7", "mon8", "mon9", "mon10", "mon11", "mon12", "t", "Lair")
 
+ds = dshape('''var * {DATE: date, AIR: float64, mon1: float64, mon2: float64,
+                      mon3: float64, mon4: float64, mon5: float64,
+                      mon6: float64, mon7: float64, mon8: float64,
+                      mon9: float64, mon10: float64, mon11: float64,
+                      mon12: float64, t: float64, Lair: float64}''')
 
-def test_resource_sas7bdat(sasfile):
+
+def test_resource_sas7bdat():
     assert isinstance(resource(test_path), SAS7BDAT)
 
 
-def test_discover_sas(sasfile):
-    ds = ", ".join(col + ": float64" for col in columns[1:])
-    expected = dshape("var * {DATE: date, " + ds + "}")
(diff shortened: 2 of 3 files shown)


into/backends/sas.py
```diff
@@ -0,0 +1,60 @@
...
+
+@convert.register(pd.DataFrame, SAS7BDAT, cost=4.0)
+def sas_to_DataFrame(s, dshape=None, **kwargs):
+    return s.to_data_frame()
+
+
+@convert.register(list, SAS7BDAT, cost=8.0)
+def sas_to_list(s, dshape=None, **kwargs):
+    s.skip_header = True
+    return list(s.readlines())
```
mrocklin: Don't need this one.  I would just depend on the `Iterator` conversion.  The conversion network will handle other transformations from `Iterator`.

into/backends/sas.py
```diff
@@ -0,0 +1,60 @@
...
+
+@convert.register(Iterator, SAS7BDAT, cost=1.0)
+def sas_to_iterator(s):
+    s.skip_header = True
+    return s.readlines()
+
+
+@append.register(sa.Table, SAS7BDAT)
+def append_sas_to_table(t, s, **kwargs):
+    append(t, sas_to_iterator(s), **kwargs)
```
mrocklin: Probably don't need this one either.

mrocklin:
This looks pretty slick.

Into has gotten smarter in recent months, so I've marked a couple of functions that probably no longer need to exist.  Otherwise things look nice.

Regarding test data.  Is there any way to use `sas7bdat` to *create* SAS files that we could use for testing?

mrocklin:
We'll also need to import it as is done here https://github.com/ContinuumIO/into/blob/master/into/__init__.py#L17-L20

and add sas7bdat to .travis.yml

into/backends/sas.py
talumbau: Yep.

into/backends/sas.py
talumbau: This one I found I needed because I needed to skip the header when iterating.

mrocklin: Hrm, but append(sa.Table, object) should itself call `sas_to_iterator` automatically.  One sec, let me grab your branch and try this out.

into/backends/sas.py
```diff
@@ -0,0 +1,60 @@
...
+import pandas as pd
+import sqlalchemy as sa
+from .sql import dshape_to_alchemy, dshape_to_table
+
+from ..append import append
+from ..convert import convert
+from ..resource import resource
+
+SAS_type_map = {'number': 'float64',
+                'string': 'string'}
```
mrocklin: Do we need to extend this?

into/backends/sas.py
mrocklin: I was able to make your tests pass while removing it.  What ends up happening is the same (we end up calling `sas_to_iterator`) but now we let `into` decide that instead of specifying it explicitly

Into is now a lot smarter about using the right sequence of `convert/append` functions for any single `into` call.  When you call `into('my-sql-table', mysasfile)` it will call `append('my-sql-table', convert(Iterator, mysasfile))` for us.

mrocklin:
Do you know how we would create an sas file from any sort of object (e.g. a Pandas DataFrame?)

talumbau:
No, but @jaredhobbs might know how to create such a SAS file.

mrocklin:
It would save us from having to hold on to binary data in the repository.  It would also be nice to craft particular datasets to test certain things.

jaredhobbs:
Unfortunately, the `sas7bdat` library is read-only. The format is closed and the only way we're able to read the data is through reverse engineering the format. While this seems to work ok for reading, there are still a lot of holes in the format that are not fully understood. Unless sas opens the format or someone can provide the missing information on the spec, there won't be any write support.

mrocklin:
Good to know.  Thanks for the work that's already there and for chiming in.

talumbau:
This work revealed the following issue (and subsequent fix):

https://bitbucket.org/jaredhobbs/sas7bdat/pull-request/5/unicode-fix-for-python-3/diff

jaredhobbs:
Thanks for the fix. I merged the change and released v2.0.3

talumbau:
same behavior in Python 2.7 and 3.x now. Work arounds for Python 2.6.

mrocklin:
Awesome.  Sorry to leave this lingering.  My bandwidth has been severely limited the last couple of days.  Should definitely have some time on Monday to finish this up.

talumbau:
@mrocklin only question I had was on the way I worked around Python 2.6. Happy to get any other feedback here too.

into/backends/sas.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+@convert.register(pd.DataFrame, SAS7BDAT, cost=4.0)
+def sas_to_DataFrame(s, dshape=None, **kwargs):
+    return s.to_data_frame()
+
+
+@convert.register(Iterator, SAS7BDAT, cost=1.0)
+def sas_to_iterator(s, **kwargs):
+    s.skip_header = True
```
mrocklin: Do we always want this?  Is there some reason why this isn't default in the `sas7bdat` library?

into/backends/tests/test_sas.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+cur_path = os.path.abspath(os.path.dirname(__file__))
+test_path = os.path.join(cur_path, 'airline.sas7bdat')
+
+
+@pytest.yield_fixture
+def sasfile():
+    with SAS7BDAT(test_path) as f:
+        yield f
```
mrocklin: I'm somewhat against using pytest idioms when we don't have to as it forces users to choose a testing tool.  In this case it seems like we might be able to get away with leaving this file open?  Happy to handle these changes myself.

into/backends/sas.py
```diff
@@ -0,0 +1,43 @@
...
+
+@discover.register(SAS7BDAT)
+def discover_sas(f, **kwargs):
+    f.skip_header = True
+    ln = next(f.readlines())
+    ds = discover(ln)
+    cols = [col.name.decode("utf-8") for col in f.header.parent.columns]
+    types = [_type.strip() for _type in str(ds).strip("()").split(',')]
+    measure = ",".join(" " + col + ":" + _type for col, _type in zip(cols, types))
+    return var * dshape("{" + measure + "}")
```
mrocklin: I was concerned about the way we discover types here.  After diving into the sas7bdat library a bit though I can't find a better way.  It looks like `f.column_types` would be good here but that seems to yield incorrect answers.  I can't find anything more robust than your method here.  I may end up changing out the string formatting for building up the DataShape programmatically.  I can do that in a separate commit though.

mrocklin:
I also removed the use of setting `skip_header`.  This seemed a bit invasive.  This did require a few changes upstream though.  Work here https://bitbucket.org/jaredhobbs/sas7bdat/pull-request/6/python-26-compatibility-and-iterator-fix/diff

Last time around @jaredhobbs was pretty speedy about merging and updating.  Here's hoping for a repeat performance :)

@talumbau I'll push my changes to fork and issue another PR.

mrocklin:
Looks like @jaredhobbs has merged and updated PyPI.  I'm good to merge with the changes in my PR to @talumbau .  Are you ok with those changes?  If so I'll go ahead and merge both into master.

talumbau:
I agree with the changes and thanks for the 2.6 fixes. Also thanks @jaredhobbs for the quick merge!

mrocklin:
This is in.  Thanks all.

PR #42, opened 2015-01-21
Minor fixes to into

cpcloud:
* add test for drop in h5py
* remove some dead code in h5py
* small spacing issue in sqlalchemy
* make sure that we can deal with fields with the string `'object'` in their name

### DIFF
--- a/into/backends/h5py.py
+++ b/into/backends/h5py.py
@@ -110,7 +110,7 @@ def varlen_dtype(dt):
     >>> r
     dtype('O')
     >>> r.metadata['vlen']
-    <type 'unicode'>
+    <type 'unicode'>  # doctest: +SKIP
     >>> dt = np.dtype([('a', 'int64'), ('b', 'object'),
     ...                ('c', [('d', 'object'), ('e', 'float64')])])
     ...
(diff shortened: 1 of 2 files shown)


into/backends/h5py.py
```diff
@@ -31,16 +33,8 @@ def discover_h5py_dataset(d):
...
 
 
-def varlen_dtype(dt):
-    """ Inject variable length string element for 'O' """
-    if "'O'" not in str(dt):
-        return dt
-    varlen = h5py.special_dtype(vlen=unicode)
-    return np.dtype(eval(str(dt).replace("'O'", 'varlen')))
-
-
```
mrocklin: Is this behavior no longer necessary?  It may be that this code is still useful but wasn't appropriately tested.

cpcloud: It was only ever returning dtype("O") so I canned it.

mrocklin: `dtype("O")` can have different metadata.  In particular, h5py accepts a dtype("O") that it specifically interprets as variable length string.

http://docs.h5py.org/en/latest/special.html
http://docs.h5py.org/en/latest/strings.html

```Python
In [1]: import h5py

In [2]: h5py.special_dtype(vlen=str)
Out[2]: dtype('O')
```

It's been a while since I dealt with this though, so my recollection is fuzzy.

cpcloud:
@mrocklin put up some more commits and a test

into/backends/h5py.py
```diff
@@ -25,22 +26,64 @@ def discover_h5py_group_file(g):
     return DataShape(Record([[k, discover(v)] for k, v in g.items()]))
 
 
+def record_dshape_replace(dshape, old, new):
```
mrocklin: Request docstring on this and the similar function below.

mrocklin:
Generally fine with this.

cpcloud:
docstrings in, merging on pass

PR #44, opened 2015-01-21
HDFS

mrocklin:
Starting an HDFS component to `into`.  Counterpoint to work done by @cpcloud on AWS

The goal is to 

1.  Create new types for remote data that allow us to cleanly leverage our existing logic
2.  Automate some migration and data loading on foreign hardware.

See discussion here:

https://github.com/ContinuumIO/into/issues/29
https://github.com/ContinuumIO/into/issues/34

```Python
In [1]: from into import HDFS, CSV, discover

In [2]: from pywebhdfs.webhdfs import PyWebHdfsClient

In [3]: hdfs = PyWebHdfsClient(host='123.123.123.123', port='14000', user_name='hdfs')

In [4]: data = HDFS(CSV)('/user/hive/warehouse/csv_test/data.csv', hdfs=hdfs)

In [5]: discover(data)
Out[5]: 
dshape("""var * {
  Name: string,
  RegistrationDate: datetime,
  ZipCode: int64,
  Consts: float64
  }""")
```

HDFS work done by @quasiben

### DIFF
--- a/into/backends/ssh.py
+++ b/into/backends/ssh.py
@@ -67,6 +67,7 @@ def resource_ssh(uri, **kwargs):
         uri = uri[len('ssh://'):]
 
     d = re.match(ssh_pattern, uri).groupdict()
+    d = dict((k, v) for k, v in d.items() if v is not None)
     path = d.pop('path')
 
     kwargs.update(d)
--- a/into/backends/ssh.py
+++ b/into/backends/ssh.py
@@ -1,0 +1,106 @@
+from __future__ import absolute_import, division, print_function
+
+import paramiko
+from contextlib import contextmanager
+from toolz import keyfilter, memoize, take
+from datashape import discover
+import re
+
+from ..utils import keywords, tmpfile
+from ..resource import resource
+
+
+class _SSH(object):
+    """ Parent class for data accessed through ssh
+
+    See ``paramiko.SSHClient.connect`` for authentication keyword arguments
+
+    Examples
+    --------
+
+    >>> from into import SSH, CSV
+    >>> SSH(CSV)('/path/to/file.csv')
+
+    Normally create through resource uris
+
+    >>> data = resource('ssh://alice@hostname:/path/to/file.csv')
+    >>> data.path
+    '/path/to/file.csv'
+    >>> data.hostname
+    'hostname'
+    """
+    def __init__(self, *args, **kwargs):
+        self.auth = keyfilter(keywords(paramiko.SSHClient.connect).__contains__,
+                              kwargs)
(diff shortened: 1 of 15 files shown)


mrocklin:
This now also includes `SSH`

into/backends/ssh.py
```diff
@@ -0,0 +1,107 @@
...
+        self.auth = keyfilter(keywords(paramiko.SSHClient.connect).__contains__,
+                              kwargs)
+        self.subtype.__init__(self, *args, **kwargs)
+
+    @memoize
+    def connect(self):
+        ssh = paramiko.SSHClient()
+        ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+        ssh.connect(**self.auth)
+        return ssh
```
mrocklin: I'm not sure what best practices are here.  Should this close itself?  Should we not memoize?  Should we keep a connection open per-host?

into/backends/ssh.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def lines(self):
+        ssh = self.connect()
+        sftp = ssh.open_sftp()
+        return sftp.file(self.path, 'r')
+
```
mrocklin: Should we have a `__str__` method here?  If so what should it look like?

into/backends/ssh.py
quasiben: I _assume_ we should not memoize and close the connection when done.   Not sure what happens when you leave it open -- the directive is `ssh.close`

into/backends/ssh.py
quasiben: suggestion for `__str__`:
<username:xxxxxxxx@hostname:port>

mrocklin: This is also a file, not really an ssh connection, so we need to put path in there somehow.  Do we want to emit this string bare or wrapped with something?

quasiben: Actually, I intended to write:
`<username:xxxxxxxx@hostname:port>`

Filename is a good point -- perhaps:
`<username:xxxxxxxx@hostname:port -- basename>`

quasiben:
This looks really great!  Tests will fail when the cluster is brought down though, no?

mrocklin:
Yes.  I have no idea how to handle testing on the HDFS code.

mrocklin:
This is getting closer to mergability.  @cpcloud any concerns?

mrocklin:
Passes what tests we have.

mrocklin:
OK, merging.

mrocklin:
@quasiben this is in.  I expect that there are lingering issues.

PR #46, opened 2015-01-26
Chunks json

mrocklin:


into/backends/json.py
```diff
@@ -192,4 +193,18 @@ def json_dumps(dt):
...
 
+@convert.register(chunks(list), chunks(JSON))
+def convert_glob_of_jsons_into_chunks_of_lists(jsons, **kwargs):
+    def _():
+        return concat(convert(chunks(list), js, **kwargs) for js in jsons)
+    return chunks(list)(_)
+
+
+@convert.register(chunks(Iterator), chunks(JSONLines))
+def convert_glob_of_jsons_into_chunks_of_lists(jsons, **kwargs):
```
cpcloud: Is there any reason why we don't have access to the first argument of the signature in `convert`? One reason I could see is that it prevents us from accidentally appending, but it also makes us have to write implementations that are the same code over and over again. If we had access to the type of the resulting conversion we could make a generic implementation that would do this. Just a thought.

mrocklin: Hrm, interesting thought.  No strong reason why it isn't there other than I found that it wasn't necessary (at the time.)

mrocklin:
This now also supports gzip loading (though not writing.)

mrocklin:
I think that I've done this in a way that is supported by all of python 2.6-3.4 and Windows/Mac/Linux but we really can't be sure until we merge and test.

mrocklin:
Merging soon if no comments.

cpcloud:
+1

PR #48, opened 2015-01-26
Allow redshift in SQL resource

cpcloud:


into/tests/test_pydot.py
```diff
@@ -6,4 +7,9 @@ def test_create_pydot_image():
         dot_graph()
     except:
         pass
-
+    finally:
+        for ext in ('pdf', 'png'):
+            try:
+                os.remove('conversions.%s' % ext)
+            except:
+                pass
```
mrocklin: FWIW I actually intended this side effect.  At the moment often use the graph for debugging.  I admit that this is a bit dirty/invasive though.

cpcloud: how about I add a little shell script to do it?

mrocklin: The `into/dot.py` file is itself a script.  We can remove the auto-generation of graphs.  I can put this into my personal workflow and stop polluting everyone elses.

cpcloud:
ok 2 merge then?

PR #50, opened 2015-01-29
Iopro support added (Issue #6)

nevermindewe:
Added support for converting from sql->ndarray via iopro if it is available.


into/backends/iopro.py
```diff
@@ -0,0 +1,159 @@
...
+    
+    if driver == "mysql":
+        #option=3 is required by mysql
+        option = "3"
+        dialect = "mysql+pyodbc"
+    elif driver == "mssql":
+        dialect = "mssql+pyodbc"
+    else:
+        raise NotImplementedError("sqlalchemy does not support the {} database "\
+                                  "with the pyodbc dialect yet.".format(driver))
```
mrocklin: I might not have understood your comments earlier about postgres, but is it not possible to use Postgres here?  Presumably we have information about where the database is and the authentication we need.  

Also, any love for `sqlite`?

mrocklin: Ah, I see, you end up creating a new SQLAlchemy engine.  Is there some lower level thing we could create instead?

into/backends/tests/test_iopro.py
```diff
@@ -0,0 +1,130 @@
...
+mysql_url = 'mysql+pymysql://continuum:continuum_test@localhost/continuum'
+pyodbc_mysql_url = 'DRIVER={mysql};SERVER=localhost;DATABASE=continuum;'\
+                   'UID=continuum;PASSWORD=continuum_test;OPTION=3;'
+
+#WARNING: pyodbc has no problem connecting to postgres,
+#  but sqlalchemy doesn't have a dialect to do so.
+#  So, we can't actually use a postgres database in this "convert" testing
+postgres_url = 'postgresql://continuum:continuum_test@localhost/continuum'
+pyodbc_postgres_url = 'DRIVER={postgresql};SERVER=localhost;DATABASE=continuum;'\
+                      'UID=continuum;PASSWORD=continuum_test;PORT=;OPTION=;'
```
mrocklin: We should probably redirect these connection strings to match those on travis.ci.  See the `test_postgres_into.py` and `test_mysql_into.py` files for connection details.

into/backends/tests/test_iopro.py
```diff
@@ -0,0 +1,130 @@
...
+    cursor = odbc.cursor()
+    cursor.execute("select tablename from pg_tables where schemaname = 'public'")
+    res = cursor.fetchall()
+    cursor.close()
+
+
+def test_iopro_sqltable_to_ndarray():
+
+    tablename = "market"
+    data = make_market_data(N=10000)
```
mrocklin: Why `N=10000`?  Perhaps this still completes in a trivial amount of time, but if not we might want to lower it.

into/backends/tests/test_iopro.py
```diff
@@ -0,0 +1,130 @@
...
+    table = populate_db(sql_engine, tablename, data)
+
+    #convert() will automatically figure out how to
+    #  go from a table to np.ndarray
+    #convert should eventually call iopro_sqltable_to_ndarray
+    res = convert(np.ndarray, table, dshape=discover(table))
+    res2 = iopro_sqltable_to_ndarray(table)
+    res3 = into(np.ndarray, table)
```
mrocklin: My understanding is that all three of these results will be the same regardless of whether or not the `iopro_sqltable_to_ndarray` function works correctly.  

Perhaps we should compare `res` against `data`?

into/backends/tests/test_iopro.py
```diff
@@ -0,0 +1,130 @@
...
+
+    assert all(res == res2)
+    assert all(res2 == res3)
+
+def test_iopro_sqlselect_to_ndarray():
+
+    tablename = "market"
+    data = make_market_data(N=10000)
+    table = populate_db(sql_engine, tablename, data)
```
mrocklin: Are we concerned about populating the same table twice?  Should the tests clean up/drop tables after they're done?

mrocklin:
I wonder if @jayvius has any thoughts on this.

into/backends/iopro.py
```diff
@@ -0,0 +1,159 @@
...
+    range.
+    """
+
+    #Remake the database engine so that iopro can use fetchsarray()
+    orig_engine = sqlselect.bind
+    iopro_db_engine = get_iopro_db_engine(orig_engine)
+
+    conn = iopro_db_engine.raw_connection()
+    cursor = conn.cursor()
+    cursor.execute(str(sqlselect))
```
mrocklin: In the case of parametrized queries consider using https://github.com/ContinuumIO/blaze/blob/master/blaze/compute/utils.py#L11

cpcloud:
we should try to get this in. i'm going to work on this today

cpcloud:
closing in favor of #148

PR #52, opened 2015-01-30
Parametrized Temp type

mrocklin:
For some remote data situations (AWS #35) (SSH/HDFS #44) it's convenient to make temporary versions of normally persistent formats like a temporary CSV file.

## Example

To this end we create a new parametrized type, `Temp`, used as follows:

```Python
In [1]: !cat foo.csv
name,balance
Alice,100
Bob,200
In [2]: from into import Temp, CSV, into

In [3]: csv = Temp(CSV)('foo.csv', delimiter=',')

In [4]: del csv

In [5]: !cat foo.csv
cat: foo.csv: No such file or directory
```

## Temp provides garbage collection on external resources

And so using `drop` and `Temp` we're able to tie persistent formats, like CSV, to the Python garbage collector.  They should be released as the object is released.

### DIFF
--- a/into/backends/csv.py
+++ b/into/backends/csv.py
@@ -20,6 +20,7 @@ from ..append import append
 from ..convert import convert, ooc_types
 from ..resource import resource
 from ..chunks import chunks
+from ..temp import Temp
 from ..numpy_dtype import dshape_to_pandas
 from .pandas import coerce_datetimes
 
@@ -104,7 +105,7 @@ def ext(path):
     return e.lstrip('.')
 
 
-@convert.register(pd.DataFrame, CSV, cost=20.0)
+@convert.register(pd.DataFrame, (Temp(CSV), CSV), cost=20.0)
 def csv_to_DataFrame(c, dshape=None, chunksize=None, nrows=None, **kwargs):
     try:
         return _csv_to_DataFrame(c, dshape=dshape,
@@ -174,7 +175,7 @@ def _csv_to_DataFrame(c, dshape=None, chunksize=None, **kwargs):
                              **kwargs2)
 
 
-@convert.register(chunks(pd.DataFrame), CSV, cost=10.0)
+@convert.register(chunks(pd.DataFrame), (Temp(CSV), CSV), cost=10.0)
 def CSV_to_chunks_of_dataframes(c, chunksize=2**20, **kwargs):
     # Load a small 1000 line DF to start
     # This helps with rapid viewing of a large CSV file
@@ -234,7 +235,7 @@ def resource_glob(uri, **kwargs):
     return chunks(type(r))(_)
 
 
-@convert.register(chunks(pd.DataFrame), chunks(CSV), cost=10.0)
+@convert.register(chunks(pd.DataFrame), (chunks(CSV), chunks(Temp(CSV))), cost=10.0)
 def convert_glob_of_csvs_to_chunks_of_dataframes(csvs, **kwargs):
     def _():
         return concat(convert(chunks(pd.DataFrame), csv, **kwargs) for csv in csvs)
--- a/into/backends/csv.py
+++ b/into/backends/csv.py
@@ -14,6 +14,7 @@ import pandas as pd
 import os
 import gzip
 import bz2
+import uuid
 
 from ..utils import keywords
 from ..append import append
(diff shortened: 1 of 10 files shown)


into/backends/csv.py
```diff
@@ -237,13 +239,20 @@ def _():
...
+@convert.register(chunks(pd.DataFrame), (chunks(CSV), chunks(Temp(CSV))), cost=10.0)
 def convert_glob_of_csvs_to_chunks_of_dataframes(csvs, **kwargs):
     def _():
         return concat(convert(chunks(pd.DataFrame), csv, **kwargs) for csv in csvs)
     return chunks(pd.DataFrame)(_)
 
 
+@convert.register(Temp(CSV), (pd.DataFrame, chunks(pd.DataFrame)))
+def convert_dataframes_to_temporary_csv(data, **kwargs):
+    fn = '.%s.csv' % uuid.uuid1()
```
cpcloud: could this also be `gzip`ped? should we handle that case?

mrocklin: Hrm, we usually specify gzipped-ness through the filename, which we no longer have.  What is a good way to specify that we want the temporary file to be gzipped in this case?

into/backends/json.py
```diff
@@ -90,12 +93,12 @@ def discover_jsonlines(j, n=10, encoding='utf-8', **kwargs):
...
+@convert.register(list, (JSON, Temp(JSON)))
 def json_to_list(j, dshape=None, **kwargs):
     return json_load(j.path, **kwargs)
 
 
-@convert.register(Iterator, JSONLines)
+@convert.register(Iterator, (JSONLines, Temp(JSONLines)))
 def json_lines_to_iterator(j, encoding='utf-8', **kwargs):
     with json_lines(j.path, encoding=encoding) as lines:
         for item in pipe(lines, filter(nonempty), map(json.loads)):
```
cpcloud: if `loads` is ever a bottleneck we can use `pandas.io.json.loads`

into/backends/ssh.py
```diff
@@ -171,3 +186,31 @@ def append_sshX_to_X(target, source, **kwargs):
...
+@convert.register(Temp(SSH(CSV)), (Temp(CSV), CSV))
+def csv_to_temp_ssh_csv(data, **kwargs):
+    fn = '.%s.csv' % uuid.uuid1()
+    target = Temp(SSH(CSV))(fn, **kwargs)
+    return append(target, data, **kwargs)
+
+
+@convert.register(Temp(JSON), (Temp(SSH(JSON)), SSH(JSON)))
+@convert.register(Temp(JSONLines), (Temp(SSH(JSONLines)), SSH(JSONLines)))
+@convert.register(Temp(CSV), (Temp(SSH(CSV)), SSH(CSV)))
```
cpcloud: nice i didn't know you could chain this

into/backends/json.py
mrocklin: Hrm, interesting

```Python
In [4]: %timeit pandas.io.json.loads('{"name": "alice", "amount": 100}')
1000000 loops, best of 3: 424 ns per loop

In [5]: import json

In [6]: %timeit json.loads('{"name": "alice", "amount": 100}')
100000 loops, best of 3: 3.15 µs per loop
```

cpcloud: pandas' version is written in C and I think the stdlib version is pure Python

into/backends/csv.py
cpcloud: i don't see any other way besides using a `compression` keyword arg

cpcloud: I suppose yet another `GZip` class might be useful

cpcloud: i'm not sure if that works with our existing "convention" method

cpcloud: You run into monstrosities like this though:

```python
into(Temp(GZip(JSONLines)), 'data.csv')
```

cpcloud: i guess that's not *so* bad

mrocklin: Don't forget to put a `Directory` of them on `HDFS`

cpcloud: it's comical how complex this could get:

```python
into(HDFS(Directory(Temp(GZip(JSONLines)))), 'data.csv')
```

"put a directory of temporary gzipped jsonlines files onto a hadoop file system"

mrocklin: I'd like to punt for the moment on gzip.  I'm inclined to use a `compression=` kwarg.

Actually, thanks to us passing all kwargs everywhere this might have already worked as is.  Sadly we don't send `**kwargs` to `DataFrame.to_csv` and, even worse, `DataFrame.to_csv` doesn't seem to support gzip writing.

cpcloud: totally ok with punting

cpcloud:
looks godo

PR #59, opened 2015-02-03
DataFrames use option types on string, float, datetime

mrocklin:
This used to fail

```Python
In [1]: from into import *

In [2]: df = pd.DataFrame({'name': ['Alice', 'Bob', 'Charlie'], 'balance':
[100, None, 300]})

In [3]: into('sqlite:///foo.db::foo2', df)
```

Now it doesn't.

cc @jiffyclub


jiffyclub:
Thanks, this fixed the issue I was having moving a DataFrame into a sqlite database.

PR #61, opened 2015-02-03
Hdfs file transfer

mrocklin:
This now works

    into('hdfs://...', 'myfile.csv')

It uses the `create_file` or `append_file` methods of the webhdfs client.

Question - I currently dump the entire data of the file into the call.  Is this a terrible idea?  Should I partition first?

@danielfrg @quasiben

### DIFF
--- a/into/backends/hdfs.py
+++ b/into/backends/hdfs.py
@@ -14,7 +14,7 @@ from collections import namedtuple
 from contextlib import contextmanager
 from .ssh import SSH, _SSH
 from .sql import metadata_of_engine, sa
-from ..utils import tmpfile, sample, ignoring
+from ..utils import tmpfile, sample, ignoring, raises
 from ..append import append
 from ..resource import resource
 from ..directory import _Directory, Directory
@@ -328,22 +328,39 @@ def append_remote_csv_to_new_table(tbl, data, dshape=None, **kwargs):
     return append(tbl2, data, **kwargs)
 
 
-@append.register(HDFS(JSONLines), JSONLines)
-@append.register(HDFS(JSON), JSON)
-@append.register(HDFS(CSV), CSV)
-def append_local_file_to_hdfs(target, source, **kwargs):
+@append.register(sa.Table, (SSH(CSV), SSH(Directory(CSV))))
+def append_remote_csv_to_table(tbl, csv, **kwargs):
     """
     Load Remote data into existing Hive table
     """
-    try:
-        target.hdfs.list_dir(target.path.lstrip('/'))
-        with open(source.path) as f:
-            # TODO: handle large files
-            target.hdfs.append_file(target.path.lstrip('/'), f)
-    except FileNotFound:
-        with open(source.path) as f:
-            # TODO: handle large files
-            target.hdfs.create_file(target.path.lstrip('/'), f)
+    path = csv.path
+    if path[0] != '/':
+        path = '/home/%s/%s' % (csv.auth['username'], csv.path)
+
+    if tbl.bind.dialect.name == 'hive':
+        statement =('LOAD DATA LOCAL INPATH "%(path)s" INTO TABLE %(tablename)s'
+                     % {'path': path, 'tablename': tbl.name})
+    else:
+        raise NotImplementedError("Don't know how to migrate csvs on remote "
+                  "disk to database of dialect %s" % tbl.engine.dialect.name)
+    with tbl.bind.connect() as conn:
+        conn.execute(statement)
+    return tbl
+
+
+@append.register(HDFS(JSONLines), JSONLines)
+@append.register(HDFS(JSON), JSON)
+@append.register(HDFS(CSV), CSV)
+def append_local_file_to_hdfs(target, source, blocksize=2**23, **kwargs):
+    if raises(FileNotFound,
+              lambda: target.hdfs.list_dir(target.path.lstrip('/'))):
+        target.hdfs.create_file(target.path.lstrip('/'), '')
+
+    with open(source.path, 'rb') as f:
+        data = f.read(blocksize)
+        while data:
+            target.hdfs.append_file(target.path.lstrip('/'), data)
+            data = f.read(blocksize)
 
     return target
 
--- a/into/backends/hdfs.py
+++ b/into/backends/hdfs.py
@@ -442,3 +442,27 @@ def resource_hdfs(uri, **kwargs):
         subtype = type(resource(path))
 
     return HDFS(subtype)(path, **kwargs)
+
+
+@append.register(HDFS(TextFile), SSH(TextFile))
+@append.register(HDFS(JSONLines), SSH(JSONLines))
+@append.register(HDFS(JSON), SSH(JSON))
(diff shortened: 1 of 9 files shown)


mrocklin:
Sending a GB file in 4k chunks sounds unpleasant

danielfrg:
It does sound unpleasant, i dont remember where i saw that.

I found this, a little bit dated but should give and idea: http://randomlydistributed.blogspot.com/2012/01/webhdfs-performance.html
So apparently ssh/scp and then `hadoop fs` is not that good

I'll try this latter and see how it works without chunks.

mrocklin:
Wow.  Those numbers are surprising.

OK, WebHDFS it is.

> I'll try this latter and see how it works without chunks.

I'll be interested in those results.

into/backends/hdfs.py
```diff
@@ -327,24 +328,24 @@ def append_remote_csv_to_new_table(tbl, data, dshape=None, **kwargs):
...
-    with tbl.bind.connect() as conn:
-        conn.execute(statement)
-    return tbl
+    try:
+        target.hdfs.list_dir(target.path.lstrip('/'))
+        with open(source.path) as f:
+            # TODO: handle large files
+            target.hdfs.append_file(target.path.lstrip('/'), f.read())
+    except FileNotFound:
+        with open(source.path) as f:
```
quasiben: You can pass in the file object -- you don't have to read it

```python
with open(source.path) as f:
    hdfs.create_file(target.path.lstrip('/'), file_data=f, overwrite=True)
```

quasiben: you can also give append_file a buffersize:
```
with open(...) as f:
    hdfs.append_file(my_file, file_data=f, buffersize=2048)
```

mrocklin: Cool.  I've added the use of just the file.  What is WebHDFS's behavior here?  Does it try to shove everything into a giant message or does it open a connection?

mrocklin:
Hive tests have recently started failing for me.  Has anything changed on the cluster?

quasiben:
Yeah -- i've been futzing with it.  I'll try and set it back to a working state

mrocklin:
No rush.  There are other things I can work on.

mrocklin:
Plus this made me improve my tests

quasiben:
ok -- should be working now

quasiben:
Actually, may be better to test again this cluster: http://54.205.229.93

into/backends/hdfs.py
quasiben: I believe it opens a connection `application/octet-stream` is in the header which indicates a binary file to be uploaded.  `requests` expects a filehandle.  I suppose we should switch to `open(..., 'rb').  Tried digging around for what webhdfs does but I can't easily find the source

danielfrg:
Couple of things:

I got this warning while importing into: `from into import into`

```
/Users/drodriguez/anaconda/envs/conda-cluster/lib/python2.7/site-packages/multipledispatch/dispatcher.py:23: AmbiguityWarning: 
Ambiguities exist in dispatched function append

The following signatures may result in ambiguous behavior:
	[HDFS(JSON), JSON], [JSON, SSH(JSON)]
	[HDFS(JSONLines), JSONLines], [JSONLines, SSH(JSONLines)]
	[HDFS(CSV), CSV], [CSV, SSH(CSV)]


Consider making the following additions:

@dispatch(HDFS(JSON), SSH(JSON))
def append(...)

@dispatch(HDFS(JSONLines), SSH(JSONLines))
def append(...)

@dispatch(HDFS(CSV), SSH(CSV))
def append(...)
  warn(warning_text(dispatcher.name, ambiguities), AmbiguityWarning)
```

I had to rename the files to have `csv` extension but the files are text files, the upload worked fine though

I had to do: `into('hdfs://hdfs@54.159.199.100:50070:/user/hdfs/100Mb.csv', '/Users/drodriguez/Downloads/100Mb.csv')` instead of `into('hdfs://hdfs@54.159.199.100:50070:/user/hdfs/100Mb.txt', '/Users/drodriguez/Downloads/100Mb.txt')`

mrocklin:
You can safely ignore the warnings for now.  They do bring up a good point though.  How should we handle loading `HDFS(.) <- SSH(.)`?

Extending the fancy uri syntax to textfiles should be easy.

Very glad to know that we can load moderately large files without being clever about it.

danielfrg:
Something else i just noticed is that if i run the command twice:

```
into('hdfs://hdfs@54.159.199.100:50070:/user/hdfs/100Mb.csv', '/Users/drodriguez/Downloads/100MB.csv')
into('hdfs://hdfs@54.159.199.100:50070:/user/hdfs/100Mb.csv', '/Users/drodriguez/Downloads/100MB.csv')
```

It will append the content so the file in HDFS will be of 200 MB, i am not sure if that is expected behavior.

danielfrg:
Some (simple) stats I got uploading from my laptop (ATX office) to the cluster (EC2).

**100 MB file**

```
%%timeit -n1 -r1
into('hdfs://hdfs@54.159.199.100:50070:/user/hdfs/100MB.csv', '/Users/drodriguez/Downloads/100MB.csv')
```
```
1 loops, best of 1: 2min 48s per loop
```

Memory measured by the OS X activity monitor was around 436 MB while uploading

**1 GB**

```
%%timeit -n1 -r1
into('hdfs://hdfs@54.159.199.100:50070:/user/hdfs/1GB.csv', '/Users/drodriguez/Downloads/1GB.csv')
```
```
1 loops, best of 1: 28min 58s per loop
```

Memory usage i saw in the activity monitor: 5.64 GB for a while then 3.78 GB until it finished.

For me the times are not really surprising, specially 1GB taking ~10 times 100MB.

Memory I am guessing is because it loads the whole file into memory, so there is something to improve there.

In any case it worked fine!

mrocklin:
OK, I'm partitioning into sets of 100k lines.  Is this too much?

@quasiben HDFS complained when I sent in bytes.  I suspect even breaking on lines is pretty fundamental to how HDFS partitions data.

danielfrg:
100k sounds good to me

mrocklin:
In regards to repeated calls to into causing appends

    into('hdfs://hdfs@54.159.199.100:50070:/user/hdfs/100Mb.csv', '/Users/drodriguez/Downloads/100MB.csv')
    into('hdfs://hdfs@54.159.199.100:50070:/user/hdfs/100Mb.csv', '/Users/drodriguez/Downloads/100MB.csv')

Yes, this was intended.  This ends up being useful when you want to put a lot of things into one thing (like many CSV files into one database table).

We should probably make this more obvious in documentation.

mrocklin:
Is there reason to implement things like the following?

    HDFS(CSV) <--> SSH(CSV)

mrocklin:
And if so, how?

mrocklin:
The ambiguity warnings you saw before should now be gone.

mrocklin:
OK, this is passing.  However I had to skip a couple of python3 tests with `Temp(SSH(CSV))` that fail intermittently (but seem totally fine when run locally)

PR #70, opened 2015-02-05
Cleanup after our tests

cpcloud:


mrocklin:
Looks good.  Also, binstar-build is now testing windows builds, so you may want to hit the test button on https://binstar.org/blaze/into/builds when merged.

PR #77, opened 2015-02-08
Fix appending to a sqlite table from a CSV with a header

cpcloud:


into/backends/sql_csv.py
```diff
@@ -25,9 +25,11 @@ def copy_sqlite(dialect, tbl, csv, **kwargs):
         abspath = abspath.replace('\\', '\\\\')
     tblname = tbl.name
     dbpath = str(tbl.bind.url).split('///')[-1]
+    delim = csv.dialect.get('delimiter', ',')
+    n = 2 if csv.has_header else 1
 
     statement = """
-     (echo '.mode csv'; echo '.import {abspath} {tblname}';) | sqlite3 {dbpath}
+     pipe=$(mktemp -t pipe) && rm -f $pipe && mkfifo -m 600 $pipe && (tail +{n} {abspath} > $pipe &) && (echo '.separator {delim}'; echo ".import $pipe {tblname}";) | sqlite3 {dbpath}
```
mrocklin: The first statement is probably better if there isn't a header.  Also we'll need to be sensitive to windows.

cpcloud: We raise an `MDNotImplementedError()` if `os.name == 'nt'`, in the execution function

into/backends/sql_csv.py
```diff
@@ -152,7 +154,9 @@ def append_csv_to_sql_table(tbl, csv, **kwargs):
     # move things to a temporary S3 bucket if we're using redshift and we
     # aren't already in S3
     if dialect == 'redshift' and not isinstance(csv, (S3(CSV), Temp(S3(CSV)))):
```
mrocklin: `isinstance(...,  S3(CSV))` should be sufficient.  `isinstance` honors inheritance.

into/backends/tests/test_sql_csv.py
```diff
@@ -56,7 +53,7 @@ def test_mysql_load():
             IGNORE 1 LINES;""" % escaped_fn)
 
 
-@skipif(os.name == 'nt')
+@skipif(os.name == 'nt', reason='SQLite copier uses posix shell features')
```
mrocklin: We might want these tests to pass on windows if there is no header.

into/backends/sql_csv.py
mrocklin: Do we still get fast loads on windows in the no-header case?

cpcloud: no, that would cause us to search for a new path, which would like be through `Iterator -> sa.Table.insert()`

into/backends/tests/test_sql_csv.py
cpcloud: i'll give it a whirl, but i don't think this was ever passing

into/backends/sql_csv.py
cpcloud: i think keeping the first statement around is just going to be a bigger maintenance burden in the future. any particular reason you want to keep it?

into/backends/sql_csv.py
cpcloud: fixed

mrocklin:
> i think keeping the first statement around is just going to be a bigger maintenance burden in the future. any particular reason you want to keep it?

If the user is on windows and if there is no header then it is useful and fast.  Could code be something like the following?

```Python
if not has_header:
    statement = ...
elif not os.name = 'nt':
    statement = ...
else:
    raise NotImplementedError()
```

cpcloud:
no the commands with be different because batch commands are different than shell commands. echoing in particular is very different

cpcloud:
i'll work out a windows solution at least for things without headers

cpcloud:
well it turns out that sqlite takes arguments so that's good. we may be able to remove some code

cpcloud:
i think is ready to go, we're passing on windows, i've xfailed the tests for csvs that have a header

mrocklin:
+1

PR #78, opened 2015-02-08
Move spark to into

cpcloud:


into/backends/sparksql.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+@convert.register(list, SparkDataFrame)
+def sparksql_dataframe_to_list(df, **kwargs):
+    return list(map(tuple, df.collect()))
+
+
+@convert.register(pd.DataFrame, SparkDataFrame)
+def sparksql_dataframe_to_pandas_dataframe(df, **kwargs):
+    return pd.DataFrame(convert(list, df, **kwargs), columns=df.columns)
```
mrocklin: Do we need this or can the network take care of it?

into/backends/sparksql.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+@append.register(SQLContext, RDD)
+def rdd_to_sqlcontext(ctx, rdd, schema=None, columns=None, **kwargs):
```
mrocklin: Should just use the `dshape=` keyword.  `schema=` is deprecated.

into/backends/sparksql.py
```diff
@@ -0,0 +1,185 @@
...
+        types = schema[0].dshapes
+        schema = dshape(Record(list(zip(columns, types))))
+    sql_schema = dshape_to_schema(schema)
+    return ctx.applySchema(rdd, sql_schema)
+
+
+@discover.register(SQLContext)
+def discover_sqlcontext(ctx):
+    dshapes = valmap(discover, get_catalog(ctx))
+    return datashape.DataShape(datashape.Record(dshapes.items()))
```
mrocklin: Might want to sort the items by key for consistency.

cpcloud:
probably should've waited to put this PR up, i just wanted to checkpoint it so that i could go back to fixing sqlite things and then come back to it after that's finished

cpcloud:
@mrocklin ready for a review whenever you get a chance

into/backends/spark.py
```diff
@@ -0,0 +1,43 @@
...
+
+@append.register(SparkContext, (list, tuple, Iterator))
+def iterable_to_spark_context(sc, seq, **kwargs):
+    return sc.parallelize(seq)
+
+
+@append.register(RDD, (list, tuple))
+def sequence_to_rdd(rdd, seq, **kwargs):
+    # Hm this seems anti-pattern-y
+    return append(rdd.context, seq)
```
mrocklin: Suggest we don't do this.  We're not really adding to the existing rdd.  This should probably just err the same way we would err if we tried to append to a tuple.

into/backends/spark.py
```diff
@@ -0,0 +1,43 @@
...
+
+from datashape import var
+
+from ..into import convert, append
+from ..core import discover
+
+
+@append.register(SparkContext, (list, tuple, Iterator))
+def iterable_to_spark_context(sc, seq, **kwargs):
+    return sc.parallelize(seq)
```
mrocklin: Suggest 

```python
@append.register(SparkContext, list)
def list_to_spark_context(sc, seq, **kwargs):
    return sc.parallelize(seq)

@append.register(SparkContext, object)
def anything_to_spark_context(sc, o, **kwargs):
    return append(sc, convert(list, o, **kwargs), **kwargs)
```

into/backends/sparksql.py
```diff
@@ -0,0 +1,207 @@
...
+    SchemaRDD = SQLContext = Dummy
+    ByteType = ShortType = IntegerType = LongType = FloatType = Dummy
+    MapType = DoubleType = StringType = BinaryType = BooleanType = Dummy
+    TimestampType = DateType = StructType = ArrayType = StructField = Dummy
+
+
+@convert.register(pd.DataFrame, SchemaRDD)
+def sparksql_dataframe_to_pandas_dataframe(rdd, dshape=None, **kwargs):
+    return pd.DataFrame(convert(list, rdd, **kwargs),
+                        columns=dshape.measure.names)
```
mrocklin: This should probably just be `convert.register(list, SchemaRDD)` and then trust the network to move the data around to a `DataFrame`.  I hope that `DataFrame <- list` already does your columns trick.

into/backends/sparksql.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+_names = ('tmp%d' % i for i in itertools.count())
+
+
+@append.register(SQLContext, (list, tuple, Iterator))
+def iterable_to_sql_context(ctx, seq, **kwargs):
+    return append(ctx, ctx._sc.parallelize(seq), **kwargs)
```
mrocklin: Same as append to SparkContext above.  Usual pattern is

```Python
@append.register(Foo, Bar)
...

@append.register(Foo, object)
def ...
    return append(myfoo, convert(Bar, myobj, **kwargs), **kwargs)
```

into/backends/sparksql.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+@append.register(SQLContext, pd.DataFrame)
+def frame_to_sqlcontext(ctx, df, **kwargs):
+    return append(ctx, convert(Iterator, df, **kwargs), **kwargs)
```
mrocklin: In my ideal world the network is as sparse as we can make it (as few functions) without sacrificing performance.  My guess is that this and the function above are redundant.

into/backends/sparksql.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+@append.register(SQLContext, RDD)
```
mrocklin: Is there any way to make this a convert statement?  Does the context of the RDD know its SQLContext?

If we can get a convert route all the way to SchemaRDD then we'll be very happy because that will connect the local world to all of the things that SchemaRDD can connect to (like `HDFS(Parquet)`) in a single function call.

mrocklin: For example is there only one SQLContext per spark context?  Perhaps we can cache that?

into/backends/sparksql.py
```diff
@@ -0,0 +1,207 @@
...
+            return dshape_to_schema(ds[0])
+    if ds in dshape_to_sparksql:
+        return dshape_to_sparksql[ds]
+    raise NotImplementedError()
+
+
+def schema_to_dshape(schema):
+    if type(schema) in sparksql_to_dshape:
+        return sparksql_to_dshape[type(schema)]
+    if isinstance(schema, ArrayType):
```
mrocklin: Does the `ArrayType` have a length?

into/backends/tests/conftest.py
```diff
@@ -0,0 +1,13 @@
...
+@pytest.fixture(scope='session')
+def sc():
+    pyspark = pytest.importorskip('pyspark')
+    return pyspark.SparkContext('local', 'into-test')
+
+
+@pytest.fixture(scope='session')
+def sqlctx(sc):
+    sparksql = pytest.importorskip('pyspark.sql')
+    return sparksql.SQLContext(sc)
```
mrocklin: This tells me that it's probably possible to convert a `PySparkRDD` to a `SchemaRDD/spark.DataFrame` which makes me pretty happy.

into/backends/sparksql.py
cpcloud: no

into/backends/sparksql.py
cpcloud: One can have multiple `SQLContext`s:

```python
In [6]: sc
Out[6]: <pyspark.context.SparkContext at 0x111c06c50>

In [7]: from pyspark.sql import SQLContext

In [8]: sql = SQLContext(sc)

In [9]: sql
Out[9]: <pyspark.sql.SQLContext at 0x111c84b10>

In [10]: sql2 = SQLContext(sc)

In [11]: sql2
Out[11]: <pyspark.sql.SQLContext at 0x111c84ad0>
```

cpcloud: `SparkContext` knows nothing about the other contexts it contains, AFAICT

cpcloud: Oh, but a `SchemaRDD` will know its `SQLContext`

cpcloud: I don't think that helps though.

into/backends/tests/conftest.py
cpcloud: Done in 09faee2

cpcloud: I'm not sure what would happen if we tried to join tables in different contexts. I suspect it will harm kittens.

mrocklin: Then perhaps we should put all into-made RDDs into a single SQLContext?  Should we cache SQLContexts based on SparkContexts?

into/backends/spark.py
```diff
@@ -0,0 +1,47 @@
...
+
+from .. import convert, append, discover
+
+
+@append.register(SparkContext, list)
+def list_to_spark_context(sc, seq, **kwargs):
+    return sc.parallelize(seq)
+
+
+@append.register((SparkContext, SQLContext), object)
```
mrocklin: For `SparkContext` I would expect the intermediate type to be an `RDD`, not a `list`

cpcloud: I'm not sure I understand what you mean. You can't add an arbitrary `RDD` to a `SparkContext`. The intermediate type must be iterable.

mrocklin: Sorry, meant `SQLContext`.  If I try to convert `into(mySQLContext, my_pandas_dataframe)` then I would expect this to go through `DataFrame -> list -> RDD -> SchemaRDD`.  I would expect the last hop to be from an `RDD` into a `SQLContext`.  This is also the `append` implementation that you have below.  

Ah, now that I see it you have two different `append` functions.  When would I prefer one over the other, e.g. what does `list -> SchemaRDD` have over `list -> RDD -> SchemaRDD`?

into/backends/tests/conftest.py
cpcloud: I'd like to post-pone doing that because I'd rather not add any extra classes right now. If you think this is critical, though, I'll do it.

mrocklin: I don't think this requires extra classes.  I think it means caching the SQLContext call in this function:

```python
+@convert.register(SchemaRDD, (RDD, PipelinedRDD))
+def rdd_to_schema_rdd(rdd, **kwargs):
+    return append(SQLContext(rdd.context), rdd, **kwargs)
```

Every time we move from pyspark to sparksql land we're making a new sqlcontext.  I'm suggesting something like the following

    SQLContext = memoize(SQLContext)

cpcloud: Ahhhh of course. It might be time for a break

into/backends/spark.py
cpcloud: That is what will happen, I've made it so that there's only a single `SQLContext` `append` function

into/backends/spark.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+@append.register(SparkContext, list)
+def list_to_spark_context(sc, seq, **kwargs):
+    return sc.parallelize(seq)
+
+
+@append.register(SparkContext, object)
+def anything_to_spark_context(sc, o, **kwargs):
+    return append(sc, convert(list, o, **kwargs), **kwargs)
```
mrocklin: If someone does the following

```python
sc = pyspark.SparkContext(...)
sql = pyspark.SQLContext(sc)

rdd = sc.parallelize(...)
srdd = into(sql, rdd)
```

Then I think that this will trigger an unnecessary collect/parallelize.

mrocklin: Oh, wait, I confused `SparkContext` with `SQLContext` again.

mrocklin: OK, what's below makes perfect sense to me.  My apologies.

mrocklin:
+1

Future fun work might be things like 

    into(my_sqlcontext, 'hdfs://hostname:myfile.json')
    into(HDFS(Parquet), my_srdd)

cpcloud:
alrighty, bombs away :bomb:

PR #86, opened 2015-02-11
Resource function for *.csv no longer requires leading text

### LINKED ISSUE #85: Importing '*.csv' into DataFrame errors with: File *.csv does not exist
A common use of `into` would be to import/convert all of the csv files in a directory. This works fine if there is a prefix to the asterix, e.g.,
`into(pd.DataFrame, 'examp*.csv')`

but errors if there is no prefix,
```
>>> into(pd.DataFrame, '*.csv')
...
IOError: File *.csv does not exist
```

mrocklin:
Before the regex was `.+\*.+` which due to the first `+`, did not
support `*.csv` because it didn't have anything leading up to the `*`.

We replace the `+` with a `*` to allow for zero leading characters

Fixes #85

### DIFF
--- a/into/backends/csv.py
+++ b/into/backends/csv.py
@@ -224,7 +224,7 @@ def resource_csv(uri, **kwargs):
 
 
 from glob import glob
-@resource.register('.+\*.+', priority=12)
+@resource.register('.*\*.+', priority=12)
 def resource_glob(uri, **kwargs):
     filenames = sorted(glob(uri))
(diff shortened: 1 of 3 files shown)


ethanwhite:
I can confirm that this fixes #85. Thanks!

mrocklin:
Great!  Merged.  As always, thank you for reporting issues.  The scope of `into` is large enough that we really rely on good reporting from the community.

PR #87, opened 2015-02-11
S3 skipping

cpcloud:


into/backends/aws.py
```diff
@@ -22,13 +20,18 @@
 from multipledispatch import MDNotImplementedError
 
 from .text import TextFile
-from ..utils import tmpfile, ext, sample
+from ..utils import tmpfile, ext, sample, keywords
+
+
+def filter_kwargs(f, kwargs):
+    return keyfilter(keywords(f).__contains__, kwargs)
```
mrocklin: Put in `utils.py`?  We do this a lot.

into/backends/sql_csv.py
```diff
@@ -6,13 +6,12 @@
...
 import sqlalchemy
 
-from multipledispatch import MDNotImplementedError
-
 from ..regex import RegexDispatcher
 from ..append import append
 from .csv import CSV, infer_header
 from ..temp import Temp
 from ..into import into
+from .aws import S3
```
mrocklin: Is this import guarded in aws.py?

into/backends/tests/test_aws.py
```diff
@@ -13,6 +13,11 @@
 
 import os
 import itertools
+import json
+try:
+    from urllib2 import urlopen
+except ImportError:
+    from urllib.request import urlopen
```
mrocklin: Could go in `compatibility.py`

into/backends/tests/test_aws.py
```diff
@@ -40,8 +44,38 @@
...
+                is_authorized = True
+            except Exception as e:
+                pytest.skip('authorization to access redshift cluster failed %s' %
+                            e)
+            else:
+                is_authorized = True
+            finally:
+                tried = True
+        else:
+            pytest.skip('authorization to access redshift cluster failed')
```
mrocklin: I'm just gonna trust you on this one :)

into/backends/sql_csv.py
cpcloud: It doesn't need to be because it's only when you instantiate it will it raise, like this:

```python
S3(CSV)('s3://mybucket/mykey.csv')
```

`S3(CSV)` won't raise

into/backends/tests/test_aws.py
cpcloud: This is kind of gross, but for testing i wanted to automate adding whatever my current ip address was to the default cluster security group. For example, if I get assigned a new ip address I don't want to have to log in to the AWS console and add my current address to the security group. This also allows us to determine whether or not someone will be able to access the cluster at all, and therefore we don't have to depend on timing out to determine whether or not we can connect.

cpcloud:
@mrocklin this is passing, any more comments before merging?

cpcloud:
actually, there's one more thing i wanted to do

cpcloud:
ok this is ready for another review, and i'll merge if everything is kosher

mrocklin:
+1

PR #94, opened 2015-02-13
SSH <- S3 with no local copies

cpcloud:


into/backends/aws.py
```diff
@@ -235,3 +237,21 @@ def remote_text_to_s3_text(a, b, **kwargs):
...
+if sys.platform != 'win32':
+    @append.register(SSH(CSV), S3(CSV))
+    def s3_to_ssh(ssh, s3, url_timeout=600, **kwargs):
+        url = s3.object.generate_url(url_timeout)
+        command = "wget '%s' -qO- >> '%s'" % (url, ssh.path)
+        conn = connect(**ssh.auth)
+        _, stdout, stderr = conn.exec_command(command)
+
+        while not stdout.channel.exit_status_ready():
+            pass
```
mrocklin: Maybe `time.sleep(...)` in order to give up control?

mrocklin: Or is there a way to block on `stdout.read()`?

into/backends/aws.py
```diff
@@ -235,3 +237,21 @@ def remote_text_to_s3_text(a, b, **kwargs):
 @append.register(HDFS(TextFile), S3(TextFile))
 def other_remote_text_to_s3_text(a, b, **kwargs):
     raise MDNotImplementedError()
+
+
+if sys.platform != 'win32':
```
mrocklin: It makes sense why this wouldn't work with a windows remote host.  Shouldn't this be usable from a windows machine though?  A windows user sshing into a linux host.

into/backends/aws.py
cpcloud: looks like this might be the ticket:

> recv_exit_status(self) method of paramiko.channel.Channel instance
    Return the exit status from the process on the server.  This is
    mostly useful for retrieving the results of an `exec_command`.
    If the command hasn't finished yet, this method will wait until
    it does, or until the channel is closed.  If no exit status is
    provided by the server, -1 is returned.

cpcloud: did a bit of source diving:

https://github.com/paramiko/paramiko/blob/master/paramiko/channel.py#L295

looks like that method blocks.

into/backends/aws.py
cpcloud: I didn't want to assume that windows users have an ssh client.

mrocklin: What if they do?  How does paramiko work in this case?

cpcloud: ok, yep. you're right. you can use paramiko to connect to a remote host on windows. i guess i shouldn't be surprised :)

cpcloud:
@mrocklin looks like we are passing. ready for round 2

cpcloud:
just remembered! we are now passing on windows on some of the previously `xfail`ing `sql_csv` tests, let me turn those off

into/backends/aws.py
```diff
@@ -235,3 +236,19 @@ def remote_text_to_s3_text(a, b, **kwargs):
 @append.register(HDFS(TextFile), S3(TextFile))
 def other_remote_text_to_s3_text(a, b, **kwargs):
     raise MDNotImplementedError()
+
+
+@append.register(_SSH, _S3)
```
mrocklin: Can you verify that this doesn't raise ambiguities?

mrocklin:
Other than the concern about ambiguities, I'm +1

into/backends/aws.py
cpcloud: It doesn't, but I'll change them to be explicit anyway

mrocklin:
+1

PR #95, opened 2015-02-13
Convert single column thing to a Series properly

cpcloud:


mrocklin:
+1.  It's a little odd because we're changing datashapes but that seems mostly harmless

cpcloud:
ok mergning

PR #97, opened 2015-02-14
JSONLines files and dirs for SparkSQL

cpcloud:


into/backends/sparksql.py
```diff
@@ -31,6 +33,19 @@ def iterable_to_sql_context(ctx, seq, **kwargs):
...
 
 
+@append.register(SQLContext, (JSONLines, Directory(JSONLines)))
+def jsonlines_to_sparksql(ctx, json, dshape=None, name=None, schema=None,
+                          samplingRatio=0.25, **kwargs):
+    # if we're passing in schema, assume that we know what we're doing and
+    # bypass any automated dshape inference
+    if dshape is not None and schema is None:
+        schema = dshape_to_schema(dshape.measure
+                                  if isrecord(dshape.measure) else dshape)
```
mrocklin: FWIW I often assume that we're given a `dshape`.  `into` does this work for us.  Also, I'd like to deprecate the `schema=` kwarg.

mrocklin: Oh, unless this is a sparksql thing and not a datashape thing.

cpcloud: Ok. I use schema explicitly here because it's an argument that `SQLContext.jsonFile` accepts.

cpcloud: i'm ok with leaving `schema` out for now.

cpcloud: in this case, one may want to let spark figure out the schema, which it will try to do if `jsonFile(schema=None)`

mrocklin: I didn't mean to scare away schema.  My last comment was "oh, unless this
is a sparksql thing"

I thought it might be a holdover from my previous code when we used to use
schema as the dshape of a record.  If a backend library uses it then it's
fine with me.

PR #99, opened 2015-02-14
S3 -> Spark and SparkSQL

cpcloud:
Adds support for loading S3 csv and JSON into Spark and SparkSQL

### DIFF
--- a/into/backends/csv.py
+++ b/into/backends/csv.py
@@ -12,9 +12,10 @@ import os
 import gzip
 import bz2
 import uuid
+from csv import Sniffer
 
 from ..compatibility import unicode
-from ..utils import keywords, ext
+from ..utils import keywords, ext, sample
 from ..append import append
 from ..convert import convert, ooc_types
 from ..resource import resource
@@ -275,6 +276,15 @@ def infer_header(csv, encoding='utf-8', **kwargs):
     return (len(df) > 0 and
             all(re.match('^\s*\D\w*\s*$', n) for n in df.columns) and
             not all(dt == 'O' for dt in df.dtypes))
+
+
+def get_delimiter(csv, length=8192):
+    """Figure out the delimiter of a CSV file.
+    """
+    sniffer = Sniffer()
+    with sample(csv, length=length) as fn:
+        with open(fn, mode='rb') as f:
+            return sniffer.sniff(f.read()).delimiter
 
 
 ooc_types.add(CSV)
--- a/into/backends/spark.py
+++ b/into/backends/spark.py
@@ -82,7 +82,6 @@ def s3_csv_to_rdd(sc, s3, dshape=None, use_unicode=False, minPartitions=None,
     # remove the first line if we have a header, assuming it's on the first
     # partition
     if infer_header(s3) if s3.has_header is None else s3.has_header:
-        return split.mapPartitionsWithIndex(lambda idx, it: (drop(1, it)
-                                                             if not idx
-                                                             else it))
+        return split.mapPartitionsWithIndex(lambda i, it:
+                                            drop(1, it) if not i else it)
     return split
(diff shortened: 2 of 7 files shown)


into/backends/sparksql.py
```diff
@@ -57,7 +65,16 @@ def rdd_to_sqlcontext(ctx, rdd, name=None, dshape=None, **kwargs):
...
     sql_schema = dshape_to_schema(dshape)
-    sdf = ctx.applySchema(rdd, sql_schema)
+
+    try:
+        sdf = ctx.applySchema(rdd, sql_schema)
+    except TypeError:
+        # TODO: gross
+        # we have to convert a tuple of strings to have typed elements
+        # this seems like it would be horribly inefficient
+        rdd = rdd.map(lambda x: convert(np.ndarray, [x], dshape=dshape).item())
```
cpcloud: i wonder if we could use `dynd` here

into/backends/csv.py
```diff
@@ -277,4 +278,13 @@ def infer_header(csv, encoding='utf-8', **kwargs):
             not all(dt == 'O' for dt in df.dtypes))
 
 
+def get_delimiter(csv, length=8192):
+    """Figure out the delimiter of a CSV file.
+    """
+    sniffer = Sniffer()
+    with sample(csv, length=length) as fn:
+        with open(fn, mode='rb') as f:
+            return sniffer.sniff(f.read()).delimiter
```
mrocklin: Suggest generalizing to `get_dialect` which returns a dict with all of the delimiter terms

into/backends/spark.py
```diff
@@ -32,6 +34,11 @@ def anything_to_spark_context(sc, o, **kwargs):
     return append(sc, convert(list, o, **kwargs), **kwargs)
 
 
+@append.register(RDD, RDD)
+def append_rdd_to_rdd(a, b, **kwargs):
+    return a.union(b)
```
mrocklin: We should be aware that this breaks conventions of `append` (which are pretty loose anyway) as an operation that mutates the left hand side.

into/backends/spark.py
```diff
@@ -49,3 +56,32 @@ def discover_rdd(rdd, n=50, **kwargs):
...
+
+
+@append.register(SparkContext, S3(CSV))
+def s3_csv_to_rdd(sc, s3, dshape=None, use_unicode=False, minPartitions=None,
+                  **kwargs):
+    """Convert an S3(CSV) file to a Spark RDD.
+
+    Notes
+    -----
+    Performs no type inference.
```
mrocklin: What would happen if I did

    into(my_sql_context, S3(CSV)(...))

into/backends/sparksql.py
mrocklin: We used to.  Also consider `rdd.mapPartitions` which will do this same operation but in bulk across a list of tuples at a time.

mrocklin: Note also that this would fail if the workers don't have into installed which may very well be the case.

mrocklin: Is SparkSQL able to make these coercions?

In the future when there is more time we could also consider make a Pure Python function that transformed a datashape into a function to do a conversion like this.  I'd want to wait until someone asked for this before spending time on it though.

into/backends/sparksql.py
```diff
@@ -46,6 +48,12 @@ def jsonlines_to_sparksql(ctx, json, dshape=None, name=None, schema=None,
     return srdd
 
 
+@append.register(SQLContext, S3(JSONLines))
```
mrocklin: This is cool.

cpcloud:
I'd like to wait to merge this after the next spark release. Their new `DataFrame` API fixes some of my database inspection hacks and has some breaking changes with out code that I'd like to address.

mrocklin:
What's the status here?  We're waiting on Spark 1.3?  Presumably this doesn't go in to the next release then?

cpcloud:
yep i don't think this will go in the next release

cpcloud:
closing. will revisit later.

PR #109, opened 2015-03-02
Access `args` instead of `message` on `NotImplementedError`.

filmor:
There is no `message` attribute on a `BaseException` instance, so the access in `discover` in `backends/sql.py:143`  raises an exception.

### DIFF
--- a/into/backends/sql.py
+++ b/into/backends/sql.py
@@ -140,7 +140,7 @@ def discover(metadata):
                 "\nSkipping.")
         except NotImplementedError as e:
             print("Blaze does not understand a SQLAlchemy type.\n"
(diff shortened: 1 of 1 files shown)


mrocklin:
Cool.  Was there a particular example where this was failing for you before?

filmor:
Yep, when accessing an Oracle database I got errors about `INTERVAL` missing a mapping (will look into that next) which raised the `NotImplementedError`.

mrocklin:
Looks like we need to map `sqlalchemy.types.INTERVAL` to `datashape.timedelta_` [here](https://github.com/ContinuumIO/into/blob/master/into/backends/sql.py#L26-L38)

PR #120, opened 2015-03-08
WIP - Blob file

mrocklin:
It would be nice to be able to move files around without understanding their file type.  This is particularly useful for the type modifiers like ssh, s3, and hdfs.

This is preliminary.  I may want to be more clever.


odo/backends/blob.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+@datashape.discover.register(File)
+def discover_file(data, **kwargs):
+    return datashape.var * datashape.bytes_
+
+
+@append.register(File, File)
+def append_file_to_file(tgt, src, **kwargs):
+    shutil.copy(src.path, tgt.path)
```
cpcloud: This will make a new file every time. Is that intended?

odo/backends/blob.py
```diff
@@ -0,0 +1,46 @@
...
+from ..convert import convert, ooc_types
+from ..append import append
+from ..drop import drop
+import os
+import shutil
+import datashape
+from multipledispatch import MDNotImplementedError
+
+
+class File(object):
```
cpcloud: FWIW the `py` library, from the same people who do `py.test` has objects like this that handle most of the operations that you're doing below. I'm not sure another dependency is in order, but it might be useful to see what they're doing over there: https://pylib.readthedocs.org/en/latest/

quasiben:
I would definitely use this for the URL routing as well

llllllllll:
If someone wants to own this please feel free to reopen.

PR #123, opened 2015-03-09
Fix escaping and encoding in mysql

cpcloud:


cpcloud:
probably going to merge this early tomorrow.

nothing here should be too controversial, i just noticed that we weren't actually running any of our mysql tests and when i got them running i uncovered a few lurking bugs

cpcloud:
mergning on pass, windows is fixed up now. there were some lingering path escaping issues

PR #128, opened 2015-03-11
Dask migration

cowlicks:
This brings in `dask/array/into.py`  and its associated tests.

I was a little confused by some of the imports that are unused in `into.py` (now `dask_array.py`). How do I know what is needed to provide an API and what is just a hanging unnecessary import? Is there a reference API for what an odo backend needs to provide?

Later I can bring in the stuff from `dask/bag/into.py`  and combine it with `dask_array.py` to make a `dask.py` backend. For now I just wanted to make sure I was on the right track.


cowlicks:
So... Stuff I didn't touch is failing. I think this is because odo's "network" is now routing some conversions through dask?

odo/backends/dask_array.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+@convert.register(Array, tuple(arrays), cost=0.01)
+def array_to_dask(x, name=None, blockshape=None, **kwargs):
+    return from_array(x, blockshape=blockshape, name=name, **kwargs)
+
+
+@convert.register(np.ndarray, Array, cost=0.5)
+def dask_to_numpy(x, **kwargs):
+    return rec_concatenate(get(x.dask, x._keys(), **kwargs))
```
mrocklin: Some of these costs probably aren't in line with other costs in `odo` (they're all done ad-hoc).  For example, this cost of 0.5 might be less than the straight `h5py.Dataset -> np.ndarray` cost, causing `odo` to route through dask even when not necessary.  This is my fault from the previous code in the dask library.

odo/backends/dask_array.py
```diff
@@ -0,0 +1,74 @@
...
+from dask.compatibility import long
+
+from ..append import append
+from ..convert import convert
+from ..resource import resource
+
+@discover.register(Array)
+def discover_dask_array(a, **kwargs):
+    block = a._get_block(*([0] * a.ndim))
+    return DataShape(*(a.shape + (discover(block).measure,)))
```
mrocklin: We not implement a `dtype` property on dask.arrays so this could be simplified.  The `datashape.from_numpy` function can be useful here.

odo/backends/dask_array.py
```diff
@@ -0,0 +1,74 @@
...
+    def resize(x, size):
+        return x.resize(size)
+except ImportError:
+    pass
+
+
+
+@convert.register(Array, tuple(arrays), cost=0.01)
+def array_to_dask(x, name=None, blockshape=None, **kwargs):
+    return from_array(x, blockshape=blockshape, name=name, **kwargs)
```
mrocklin: Many of the errors arise because this function expects a non-None blockshape but isn't getting it.  It should probably check for this case and then raise `NotImplementedError` which will cause `odo` to reroute around this edge.

cowlicks: That should be raised in dask by the `dask.core.from_array` function, not here. Right?

mrocklin: That's valid.  from_array might not raise NotImplementedError though.  More
likely ValueError.  The network mechanism in odo only skips on
NotImplementedError and raises on all others.

odo/backends/dask_array.py
cowlicks: Okay, I'm not very familiar with the dask algorithms, or the other odo backends. So it is kind of a hard call for me to make.

Are there plans to actually measure all these in the future? That could be fun.

mrocklin: I think that that would be a fascinating way to study the performance of the ecosystem.  There are probably a few things like that that we're well poised to do with odo/blaze that we haven't touched at all yet.

For the immediate purpose of changing the behavior of odo my intuition says that there are probably a few improvements to be made, but nothing groundbreaking.  The graph is sufficiently sparse that there are only usually a few sensible routes and relative costs between them usually differ by orders of magnitude, making the right choice fairly obvious (usually).  There have only been a handful of cases where we've been highly sensitive to algorithm cost.  It would be interesting to see how well we do with an unweighted graph; I'll bet it's not terrible.

For the moment I would just make sure that the 

    cost(np <- dask.array <- h5py.Dataset) > cost(np <- h5py.Dataset)

Probably this cost should be bumped up to 2 or something.

cowlicks:
I'm wrestling with my dev environment right now, so I'm not sure if this is going to work. My tests are hanging on `test_sqlite_into.py::test_csv_with_header` and `test_sql_csv.py::test_into_sqlite_with_header` I might have some dependency issue but I'm not sure what is going on yet. Suggestions welcome.

odo/backends/tests/test_dask.py
```diff
@@ -0,0 +1,72 @@
+from __future__ import absolute_import, division, print_function
+
+from odo.backends.dask import append, Array, merge
+from dask.array.core import insert_to_ooc
+from dask import core
+from odo import convert, into
+from odo.utils import tmpfile
+import numpy as np
+import bcolz
+
```
mrocklin: This should skip if dask is not installed 

```python
import pytest
pytest.importorskip('dask')
```

mrocklin:
Are you on windows perhaps?  SQLite has some issues there at times.

In any event that sounds like a usability error on our side.  Can you report it as a separate issue?

mrocklin:
This is in.  Thanks @cowlicks !

PR #130, opened 2015-03-11
Hive

mrocklin:
Some updates for Hive

1.  Clean up ssh-hive tests to be less stateful
2.  Handle CSV dialects from directories
3.  Reduce the amount of full-database reflection that happens (this is expensive)
4.  Accept stored_as= and external= kwargs to create non-textfile hive tables

TODO: insert syntax in PyHive doesn't work

### DIFF
--- a/docs/source/hive.rst
+++ b/docs/source/hive.rst
@@ -1,0 +1,94 @@
+Hive Metastore
+==============
+
+The Hive metastore relates SQL metadata to files on the Hadoop File System
+(HDFS_).  It is similar to a SQL database in that it contains information about
+SQL tables but dissimilar in that data isn't stored in Hive but remains
+ordinary files on HDFS.
+
+Odo interacts with Hive mostly through ``sqlalchemy`` and also with a bit
+of custom code due to its peculiarities.
+
+URIs
+----
+
+Hive uris match exactly SQLAlchemy connection strings with the ``hive://``
+protocol.  Additionally, Impala, another SQL database on HDFS can also connect
+to the same tables.
+::
+
+    hive://hostname
+    hive://user@hostname:port/database-name
+    hive://user@hostname:port/database-name::table-name
+
+    impala://hostname::table-name
+
+Additionally you should probably inspect docs on HDFS_ due to the tight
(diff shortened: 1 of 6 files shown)


odo/backends/hdfs.py
```diff
@@ -252,17 +281,24 @@ def create_hive_statement(tbl_name, dshape, path=None, table_type='',
     # Column names and types from datashape
     ds = dshape or discover(data)
```
cpcloud: it looks like there's an undefined variable `data` here

odo/backends/sql.py
```diff
@@ -309,6 +309,18 @@ def append_anything_to_sql_Table(t, c, **kwargs):
 def append_anything_to_sql_Table(t, o, **kwargs):
     return append(t, convert(Iterator, o, **kwargs), **kwargs)
 
+@append.register(sa.Table, sa.Table)
```
cpcloud: should there be a test that this fails when trying to insert, say, a sqlite table into a hive table?

cpcloud: actually, it looks like that might actually *pass*

docs/source/hive.rst
```diff
@@ -0,0 +1,94 @@
...
+
+    Hive -> Hive
+
+And as with all SQL systems through SQLAlchemy we can convert a Hive table to a
+Python Iterator, though this is somewhat slow::
+
+    Hive -> Iterator
+
+
+Impala
```
cpcloud: I wonder if a small list of technologies that odo handles that work with hive might be nice.

mrocklin: I'm not familiar enough with spark+hive+odo interactions.  Do we have this?

odo/backends/sql.py
mrocklin: This check occurs just below in the append Select to Table call

odo/backends/hdfs.py
mrocklin: Fixed

mrocklin:
Other comments?  I'd like to merge this for the release.

PR #131, opened 2015-03-12
Fix a few spark things

cpcloud:


cpcloud:
nothing to write home about here, just removing the implicit spark dependency on `paramiko` via `HDFS`

cpcloud:
merging in a few minutes

PR #138, opened 2015-03-12
Remove superfluous "into" from several backend tests.

cowlicks:
Several test modules were still named `test_*_into.py`. I removed the `_into`.


mrocklin:
Thanks!

Issue #137, opened 2015-03-12
sqlite and csv `with_header` tests hanging. 


cowlicks:
I'm on debian testing/jessie/8.

In  the odo repository on master I make a new conda environment, in my `.condarc` I have python, ipython, and pip as default packages
```bash
$ conda create -n please-work pytest  # need pytest to run the test suite
...
$ source activate please-work
...
(please-work) $ conda install --file requirements.txt
...
(please-work) $ python setup.py install 
...
(please-work)$ conda list
# packages in environment at /home/blake/anaconda/envs/please-work:
#
datashape                 0.4.3                np19py27_0  
dateutil                  2.1                      py27_2  
decorator                 3.4.0                    py27_0  
ipython                   3.0.0                    py27_0  
multipledispatch          0.4.7                    py27_0  
networkx                  1.9.1                    py27_0  
numpy                     1.9.2                    py27_0  
odo                       0.3.0                     <pip>
openssl                   1.0.1k                        1  
pandas                    0.15.2               np19py27_0  
pip                       6.0.8                    py27_0  
py                        1.4.26                   py27_0  
pytest                    2.6.4                    py27_0  
python                    2.7.9                         2  
python-dateutil           1.5                       <pip>
pytz                      2014.9                   py27_0  
readline                  6.2                           2  
scipy                     0.15.1               np19py27_0  
setuptools                14.0                     py27_0  
six                       1.9.0                    py27_0  
sqlite                    3.8.4.1                       0  
system                    5.8                           1  
tk                        8.5.15                        0  
toolz                     0.7.1                    py27_0  
zlib                      1.2.8                         0  
```
Issue number 1: test suite depends on bcolz, h5py, and sqlalchemy
```bash
(please-work) $ py.test odo --verbose
... 
==================================== ERRORS ====================================
______________ ERROR collecting odo/backends/tests/test_bcolz.py _______________
odo/backends/tests/test_bcolz.py:3: in <module>
    from odo.backends.bcolz import (append, convert, ctable, carray, resource,
odo/backends/bcolz.py:4: in <module>
    from bcolz import ctable, carray
E   ImportError: No module named bcolz
_______________ ERROR collecting odo/backends/tests/test_h5py.py _______________
odo/backends/tests/test_h5py.py:3: in <module>
    import h5py
E   ImportError: No module named h5py
_______________ ERROR collecting odo/backends/tests/test_sql.py ________________
odo/backends/tests/test_sql.py:6: in <module>
[...]

PR #140, opened 2015-03-12
Skip test if we don't have bcolz, h5py or sqlalchemy

cowlicks:
This was brought up in issue #137 now test suite should pass with solely the things in requirements.txt


mrocklin:
Looks good to me.  Merging once tests pass.

PR #141, opened 2015-03-12
Dask bag migration

cowlicks:
Brings in stuff from `dask.bag.into` and its tests. I still need to add `costs` to the conversions here.

### DIFF
--- a/odo/backends/tests/test_dask.py
+++ b/odo/backends/tests/test_dask.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import, division, print_function
 
 import pytest
 pytest.importorskip('dask')
+pytest.importorskip('dask.bag')
 
 from odo.backends.dask import append, merge
 from dask.array.core import insert_to_ooc, Array
--- a/odo/backends/tests/test_dask.py
+++ b/odo/backends/tests/test_dask.py
@@ -3,14 +3,17 @@ from __future__ import absolute_import, division, print_function
 import pytest
 pytest.importorskip('dask')
 
-from odo.backends.dask import append, Array, merge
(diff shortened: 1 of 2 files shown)


odo/backends/tests/test_dask.py
```diff
@@ -3,14 +3,17 @@
 import pytest
 pytest.importorskip('dask')
```
mrocklin: Should also skip on `dask.bag`.  It has some dependencies that aren't explicitly stated in dask proper.

mrocklin:
This seems fine to me.

odo/backends/tests/test_dask.py
```diff
@@ -2,15 +2,19 @@
 
 import pytest
 pytest.importorskip('dask')
+pytest.importorskip('dask.bag')
```
cowlicks: @mrocklin like this? I'm not sure what the build env would look like to cause that.

mrocklin: Yes, like that.

At the moment `dask` does not list all of the dependencies of its subprojects.  If one `pip` or `conda` installs dask you don't necessarily get all of `dill`, `numpy`, `bcolz` required for each of `dask.bag`, `dask.array`, `dask.frame` respectively.  Importing dask doesn't necessarily import all of the subprojects though.  Users will get the `ImportError` only if they try to import a subproject for which their environment does not contain the necessary dependencies.

cpcloud:
@mrocklin @cowlicks is this ready for merging?

mrocklin:
+1

cowlicks:
Thanks!

PR #147, opened 2015-03-19
Add interval types to odo's sql backend

cpcloud:


mrocklin:
If we get a different kind of timestamp does this fail?

cpcloud:
do you mean a different unit?

mrocklin:
And what about for intervals with different parameters?  Parameterized types tend to need to be handled more explicitly in dshape_to_alchemy

cpcloud:
yep this is more of a WIP

mrocklin:
Ah, sorry to fire early.

cpcloud:
no problem, i'd rather have the feedback earlier than later. fire at will :gun:!

cpcloud:
@mrocklin can you review this when you get a chance? i need this for other work, not a huge rush ... but whenever you get a chance

mrocklin:
Will take a look

odo/backends/sql.py
```diff
@@ -23,46 +24,88 @@
...
+    6: 'us',
+    9: 'ns'
+}
 
 
 @discover.register(sa.sql.type_api.TypeEngine)
 def discover_typeengine(typ):
+    if isinstance(typ, sa.types.Interval):
+        if not (typ.second_precision or typ.day_precision):
+            return datashape.TimeDelta(units='us')
```
mrocklin: Is there a standard in SQL for this?

mrocklin:
+1

cpcloud:
@mrocklin do you mind if i add a parameterized test for constructing all the different frequencies?

mrocklin:
Go for it.  Your recent productivity sufficiently trumps mine so that I think your preferences should hold sway.

odo/backends/sql.py
cpcloud: I just pulled up the SQL2003 standard (I couldn't easily get access to 2008 or 2011) and there's only a default precision for non leading fields. For example if i specify a `day interval`, then by default the rest of the fields (`hour`, `minute`, and `second`) all have a precision of two integers (in the case of `second` this is two integers before the decimal point, since `second` intervals can be floating point values)

cpcloud: the default frequency *could* be seconds, but in SQL it doesn't make sense to define a field with an interval type and not give it a frequency.

cpcloud: sqlalchemy seems to allow you to construct this kind of column without a precision though, let me see if there's something else going on

cpcloud: hm. seems i was mistaken ...

postgres:

```
pcloud=# create table tb (x interval);
CREATE TABLE
pcloud=# insert into tb values('1 day');
INSERT 0 1
pcloud=# insert into tb values('2 days 00:00');
INSERT 0 1
pcloud=# insert into tb values('2 days 00:00:00.2309235');
INSERT 0 1
pcloud=# select * from tb;
           x
------------------------
 1 day
 2 days
 2 days 00:00:00.230924
```

seems that the default precision for postgres is 6 decimal places

mysql apparently doesn't have an interval type

cpcloud:
Going to look into a few databases to see what happens when I create tables with this branch.

cpcloud:
going to give this another half day. probably merging tomorrow

cpcloud:
merging, i'm decently happy with this

PR #150, opened 2015-03-23
Make sql<-csv converters robust to absence of sqlite3

mrocklin:
We've done this for Windows in the past.  Now we extend the
solution to other operating systems.

I honestly don't know how to test this well.  Perhaps the Anaconda
build system can help here.

cc @asmeurer

### DIFF
--- a/odo/backends/sql_csv.py
+++ b/odo/backends/sql_csv.py
@@ -53,8 +53,9 @@ def execute_copy_sqlite(dialect, engine, statement):
     ps = subprocess.Popen(statement, shell=True,
                           stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
     stdout, stderr = ps.communicate()
-    # Windows sometimes doesn't have sqlite3.exe
-    if os.name == 'nt' and (stderr or 'not recognized' in str(stdout)):
+    # If we don't have sqlite3.exe
+    if any(b and a in b.lower() for a in ['not recognized', 'not found']
+                                for b in [stdout, stderr]):
         raise MDNotImplementedError(stderr)
     return stdout
 
--- a/odo/backends/sql_csv.py
+++ b/odo/backends/sql_csv.py
@@ -4,6 +4,7 @@ import os
 import sys
 import re
 import subprocess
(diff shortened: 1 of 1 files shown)


cpcloud:
Is there no way we can list sqlite3 as a dependency only for windows?

asmeurer:
I would rewrite this code to not use shell=True (can't you do this processing in Python?), and use `distutils.spawn.find_executable` or something similar to see if sqlite3 is there or not.

cpcloud:
That way if someone installs blaze via conda then we can ensure that things are there.

asmeurer:
Note that this problem will go away for conda based installations of odo because we are rebuilding the sqlite package with the sqlite3 binary (and sqlite is a dependency of Python itself, so it will always be there).

cpcloud:
What versions of Python don't include SQLite?

cpcloud:
Why does this need to be generalized? Is there an OS other than windows that has a python installation without SQLite?

asmeurer:
For whatever reason, my stock ubuntu doesn't have the sqlite3 executable (but the system Python does seem to have the sqlite3 library). 

I don't know what the state of things would be on Windows. This issue originally came up on Ubuntu.

mrocklin:
> I would rewrite this code to not use shell=True (can't you do this processing in Python?), and use distutils.spawn.find_executable or something similar to see if sqlite3 is there or not.

Now using `distutils.spawn.find_executable`.  I'm pretty ignorant on the best approach here.  Quite happy to learn of better ways.

asmeurer:
The better approach is to avoid using `shell=True`. Split the command in a list and pass it to subprocess. `shell=True` leads to pretty bad shell injection vulnerabilities (if someone can use an arbitrary string as one of the variables, then they can trivially execute arbitrary shell commands). I would also replace all the shell piping Python (or at least separate subprocess calls). It's much easier to maintain Python than a shell oneliner. 

This is independent of this fix, though. I would just test that this works (remove sqlite from your PATH and run the tests).

mrocklin:
Verified that this works when sqlite3 is missing.  Merging.

PR #152, opened 2015-03-26
Fix unicode borking on binstar by setting locale envars

cpcloud:


cpcloud:
green across the board https://binstar.org/cpcloud/odo/builds/23, merging

PR #154, opened 2015-03-28
Discover on selectables

cpcloud:
adds the ability to discover on general `sa.sql.Selectable`s

### DIFF
--- a/odo/backends/sql.py
+++ b/odo/backends/sql.py
@@ -11,6 +11,7 @@ from datashape.dispatch import dispatch
 from datetime import datetime, date
 import datashape
 from toolz import partition_all, keyfilter, first, pluck, memoize, map
+from toolz import identity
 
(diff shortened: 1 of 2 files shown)


cpcloud:
in case anyone is wondering whether `Selectable` is guaranteed to have columns, the answer is yes.
Here's the blurb from the top of docs:

> The term “selectable” refers to any object that rows can be selected from; in SQLAlchemy, these objects descend from FromClause and their distinguishing feature is their FromClause.c attribute, which is a namespace of all the columns contained within the FROM clause (these elements are themselves ColumnElement subclasses).

cpcloud:
hm it appears that `FromClause` is actually the class that's guaranteed to have a `columns` attribute. Fixing.

cpcloud:
blaze passes with this ... merging

PR #155, opened 2015-03-29
Fetchall

cpcloud:


odo/backends/sql.py
```diff
@@ -321,7 +320,7 @@ def select_to_base(sel, dshape=None, **kwargs):
     engine = sel.bind  # TODO: get engine from select
 
     with engine.connect() as conn:
-        result = conn.execute(sel)
+        result = conn.execute(sel).fetchall()
```
mrocklin: Does this affect laziness at all?

cpcloud: Hm. Laziness *does* have the benefit of not blowing up memory, but without `fetchall` sqlalchemy can and often does hit the server with every call to `fetchone`. This of course depends on the dialect. I like laziness but hitting the server every single time for, say, a 1000 row result set is crazy. A happy medium might be to do something along the lines of:

```python
result = conn.execute(sel)
chunksize = 1000

def execute(conn, sel, chunksize=1000):
    result = conn.execute(sel)
    while True:
        try:
            yield result.fetchmany(size=chunksize)
        except StopIteration:
            return

for row in concat(execute(conn, sel)):
    yield row
```

mrocklin: Batching sounds good.  I'm surprised that sqlalchemy doesn't do this
already.

cpcloud:
closing in favor of a new pr that will have batching

PR #162, opened 2015-04-02
More general selectable conversion

cpcloud:
Sorry for the churn.

Implemented a more general and robust way of doing this


cpcloud:
Merging soon if no comment. Testing blaze locally now to make sure nothing breaks.

PR #167, opened 2015-04-08
Remote URL Backend

quasiben:
PR handles downloading remote data accessible from the following protocols:

- `HTTP`
- `HTTPS`
- `FTP`

PR #168, opened 2015-04-17
URL on top of #167

cpcloud:
adds more robust encoding and removes duplicate newlines from a text file when generating a sample


cpcloud:
Actually, I don't think the `TextFile` support here is necessary. It's always discovered as `var * string`

cpcloud:
@quasiben put tests in that pass when `URL(TextFile)` isn't in the signatures inside of `sample` and `discover` in `url.py`

cpcloud:
I also think this is a backend that will work without any deps, because we're only using stdlib stuff.

cpcloud:
hm maybe not. we use `csv` in it

cpcloud:
@quasiben you're next in the PR queue after #173 -> #172 -> #168

quasiben:
:)

cpcloud:
merging soon, just want to confirm that current binstar builds are passing, then will merge

PR #171, opened 2015-04-18
Restrict sqlalchemy to lower than 1.0.0

### LINKED ISSUE #169: latest odo builds hanging
tested this against a previously passing build and still hangs. likely a dependency issue.

i think the only thing major that changed was sqlalchemy 1.0.0, so that's probably the culprit.

cpcloud:
closes #169


cpcloud:
closing in favor of #173

PR #173, opened 2015-04-18
Versioneer and proper build and test of many optional deps in conda.recipe

### LINKED ISSUE #169: latest odo builds hanging

cpcloud:
closes #169


cpcloud:
after this PR we'll have:

- [ ] support for sqlalchemy 1.0.0
- [ ] proper versioning, and we *never* have to touch `__version__` again

cpcloud:
Things we lose in the process

- [x] setuptools support. only distutils plays nicely with versioneer. imo whatever we lose by not using setuptools is worth never having to deal with versioning except when we want to tag things in git

cpcloud:
@mrocklin ok with this?

cpcloud:
merging soon if no comment

cpcloud:
merging and will fix binstar builds if necessary

cowlicks:
No setuptools mean no `python setup.py develop` :(

cpcloud:
Yeah. Is there a way we could make our own develop command? You can make custom commands with the cmdclass argument.

bryevdv:
Bokeh doesn't use setuptools either, it's easy enough to write and remove your own `.pth` files:

https://github.com/bokeh/bokeh/blob/master/setup.py#L446

cpcloud:
Cool. @cowlicks care to do a PR for this? I can get to it possibly Sunday night or next week if you don't have time.

cowlicks:
Sure, I'll be able to get to it soon.

cowlicks:
I think this is breaking some stuff in blazes travis-ci builds. Before in blaze we would do:
```
pip install git+http://url-to/datashape
pip install git+http://url-to/odo
```
but before this the odo install would recognize the latest datashape is installed and leave it alone. Odo requires `datashape >= 0.4.4` which the master branch of datashape should satisfy. But now odo thinks the masterbranch is too old and downloads the 0.4.4 release of datashape, which doesn't include [this change](https://github.com/ContinuumIO/datashape/commit/1790180cc23bf4882b4bb800b6046552e9ed7091). Which blaze depends on. So some PR's there are breaking.

Failing blaze PRs: [#1041](https://github.com/ContinuumIO/blaze/pull/1041), [#1040](https://github.com/ContinuumIO/blaze/pull/1040), [#1035](https://github.com/ContinuumIO/blaze/pull/1035), [#1029](https://github.com/ContinuumIO/blaze/pull/1029)

cpcloud:
I think the solution here is to use conda  and the blaze channel for testing which can support proper development version numbers.

cowlicks:
I don't think blaze tests will still fail with the version in the blaze channel. Also, odo still overrides the datashape installed by `conda install -c blaze datashape`.

cpcloud:
I don't think they'll fail for the same reason. We should use both odo and data shape from the blaze channel and not pip install them. That will not override anything

PR #192, opened 2015-05-07
Refactor CSV has_header inference

cpcloud:


cpcloud:
@mrocklin this could use a bit of review when you get a chance

cpcloud:
things are passing on all win64 py27 and py34 as well as osx py27 and py34

odo/backends/csv.py
```diff
@@ -39,7 +43,36 @@ def alias(key):
...
+    try:
+        yield f
+    finally:
+        f.close()
+
+
+def infer_header(path, nbytes=10000, encoding='utf-8', **kwargs):
+    with open_file(path, 'rb') as f:
+        raw = f.read(nbytes)
+    return csv.Sniffer().has_header(raw if PY26 else raw.decode(encoding))
```
mrocklin: I was burned sufficiently often by windows and py2/3 differences to avoid the csv module alltogether in favor of pandas.read_csv.  If you've handled that then that sounds good.

Curious, what triggered the move away here?

cpcloud: It wasn't working automatically for tab separated files

PR #199, opened 2015-05-15
More DB -> CSV fixes and some additions

cpcloud:
- [x] pass encoding to sniffer
- [x] test that we can handle headers with quoted field names
- [x] use transactions with `begin` when writing to csv
- [x] Output CSV files always have a header unless we're using MySQL, so change that attr when returning the `CSV` instance.
- [x] test that we can handle outputting `NULL` values to CSV from postgres
- [x] that that we can read in files with a different encoding


cpcloud:
passing on windows. merging

PR #221, opened 2015-06-05
ENH: Parse schema name in sql uri

llllllllll:
I am not sure about the best practices of testing, or if this code live in the most correct spot.

### DIFF
--- a/odo/backends/sql.py
+++ b/odo/backends/sql.py
@@ -441,7 +441,7 @@ def resource_sql(uri, *args, **kwargs):
         table_name, args = args[0], args[1:]
         try:
             schema_name, table_name = table_name.split('.', 1)
-        except:
+        except ValueError:
             schema_name = None
 
         metadata = metadata_of_engine(engine)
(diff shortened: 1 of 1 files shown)


llllllllll:
status?

odo/backends/sql.py
```diff
@@ -431,7 +439,15 @@ def resource_sql(uri, *args, **kwargs):
     # we were also given a table name
     if args and isinstance(args[0], str):
         table_name, args = args[0], args[1:]
+        try:
+            schema_name, table_name = table_name.split('.', 1)
+        except:
```
cpcloud: Can you catch only the exception that raised here?

llllllllll: sorry about that, sure thing.

cpcloud:
I think we should dispatch on the dialect here. I'll show an example tomorrow.

llllllllll:
kk

llllllllll:
Closing in favor of #223

PR #222, opened 2015-06-08
BUG: Replace usages of table.name with table.key.

llllllllll:
If a table is inside of a schema, for example: 'schema.name', then the
table's name is 'name'; however, using the name in DDL will cause errors
because you must use the full key, 'schema.name'.

### DIFF
--- a/odo/backends/sql.py
+++ b/odo/backends/sql.py
@@ -413,7 +413,8 @@ def append_select_statement_to_sql_Table(t, o, **kwargs):
     if not o.bind == t.bind:
         return append(t, convert(Iterator, o, **kwargs), **kwargs)
 
-    assert o.bind.has_table(t.key), 'tables must come from the same database'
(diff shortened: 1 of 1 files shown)


odo/backends/sql.py
```diff
@@ -413,7 +413,7 @@ def append_select_statement_to_sql_Table(t, o, **kwargs):
     if not o.bind == t.bind:
         return append(t, convert(Iterator, o, **kwargs), **kwargs)
 
-    assert o.bind.has_table(t.name), 'tables must come from the same database'
+    assert o.bind.has_table(t.key), 'tables must come from the same database'
```
llllllllll: So I thought that I would solve an issue I was having here by making this change; however, it did not work.

This assertion is catching false positives and I am not positive why.

llllllllll: It appears I had already written a workaround in my code somewhere else.

llllllllll:
Closing in favor of #223

PR #223, opened 2015-06-09
Allow schema to be passed in resource

cpcloud:


odo/backends/sql.py
```diff
@@ -422,21 +427,52 @@ def append_select_statement_to_sql_Table(t, o, **kwargs):
     return t
 
 
-@resource.register('(.*sql.*|oracle|redshift)(\+\w+)?://.+')
+def create_schema(ddl, target, bind, tables=None, state=None, checkfirst=None,
```
llllllllll: This is not `create_schema` but `has_schema`, no?

cpcloud: Good call. Will change when back from lunch.

cpcloud:
@llllllllll Merge-able?

llllllllll:
lgtm

PR #234, opened 2015-06-20
missing }

gwulfs:


llllllllll:
Thanks for the fix, let me just see what is up with the build failure before merging; however, I am not sure how this would trigger the failure.

cpcloud:
Looks like a new release of pytables broke something in pandas

cpcloud:
Actually this was our fault, we have true division enabled and a chunksize was getting passed as `2.0` and casting to long when slicing a pytables `Table` was removed in the latest pytables. Fixed in #236

cpcloud:
@gwulfs Can you pull from master and push up again to rerun against the fix i put in in #236

PR #237, opened 2015-06-20
missing }, build fail

gwulfs:
trying this again


cpcloud:
Looks good to me! Merging. @gwulfs Thanks for the fix!

PR #242, opened 2015-06-22
don't prepend schema name when discovering the shape of a `MetaData` object.

llllllllll:
If I discover `MetaData(url, schema='my_schema')`, then the table names should not also be prefixed with: `my_schema.`


cpcloud:
@llllllllll does this need to be merged before ContinuumIO/blaze#1145?

llllllllll:
Err, I think they are independently tested but need to both be merged

PR #243, opened 2015-06-25
Fixes gzip resource dispatching on JSON discovery

cpcloud:


cpcloud:
passing on windows too

PR #248, opened 2015-07-01
Fix appending to h5py from other h5py

### LINKED ISSUE #218: odo-ing to an existing HDF5 dataset yields TypeError
Doing:

```python
from odo import odo
odo('file_1.hdf5::/mydataset','file_combined.hdf5::/mydataset')
```

appears to work only if file_combined.hdf5::/mydataset does not already exist.

Otherwise, it yields a TypeError. For example:

[...]

```
/usr/local/anaconda/anaconda2/lib/python2.7/site-packages/h5py/_hl/group.pyc in require_dataset(self, name, shape, dtype, exact, **kwds)
    120 
    121         if not shape == dset.shape:
--> 122             raise TypeError("Shapes do not match (existing %s vs new %s)" % (dset.shape, shape))
    123 
    124         if exact:

TypeError: Shapes do not match (existing (161793,) vs new (0,))
```

As is, this makes it impossible to concatenate multiple HDF5 datasets with odo.

cpcloud:
closes #218

### DIFF
--- a/odo/backends/h5py.py
+++ b/odo/backends/h5py.py
@@ -231,15 +231,11 @@ def resource_h5py(uri, datapath=None, dshape=None, expected_dshape=None,
         uri = uri[len('h5py://'):]
     f = h5py.File(uri)
     olddatapath = datapath
-    if datapath is not None:
-        try:
-            old_dset = f[datapath]
-        except KeyError:
-            pass
-        else:
-            if expected_dshape is not None:
-                dshape = expected_dshape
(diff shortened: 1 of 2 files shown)


cpcloud:
@mrocklin I remember that you had some dask code that was append to HDF5 files using odo. Can you confirm that your code still works with this change?

mrocklin:
I suspect that that code is no longer active.  Dask tends to manage hdf5 on its own now.

Issue #251, opened 2015-07-01
adding a few simple test cases


ywang007:
Hi, I would like to add a few simple tests for a couple files, namely:
append.py
utils.py

cpcloud:
@ywang007 great! what did you have in mind? do you need help with something in particular?

ywang007:
It's really more for me to get familiar with the github workflow. :)
I've made the change, raising the PR now.

cpcloud:
closed by #252

cpcloud:
@ywang007 useful github tip:

when you write

```
closes #some-issue-number
```
in the main message of the pull request, github will automagically closed the referenced issue number when the PR is merged

ywang007:
@cpcloud , cool. i'll try that next time. thanks.

PR #252, opened 2015-07-01
adding simple test cases

ywang007:
Issue #251: I added a few test cases. Please check and advise if they can be pulled in.

### DIFF
--- a/odo/tests/test_append.py
+++ b/odo/tests/test_append.py
@@ -12,3 +12,16 @@ def test_append_list_to_set():
     s = set([1, 2, 3])
     append(s, [4, 5, 6])
     assert s == set([1, 2, 3, 4, 5, 6])
+
+
+def test_append_set_to_list():
+    s = set([3, 4, 5])
+    l = [1, 2, 3]
+    append(l, s)
+    assert l == [1, 2, 3, 3, 4, 5]
+
+
+def test_append_tuple_to_set():
+    s = set([1, 2, 3])
+    append(s, (3, 4, 5))
+    assert s == set([1, 2, 3, 4, 5])
--- a/odo/tests/test_append.py
+++ b/odo/tests/test_append.py
@@ -18,7 +18,7 @@ def test_append_set_to_list():
     s = set([3, 4, 5])
     lst = [1, 2, 3]
     append(lst, s)
-    assert lst == [1, 2, 3, 3, 4, 5]
+    assert sorted(lst) == [1, 2, 3, 3, 4, 5]
 
 
 def test_append_tuple_to_set():
--- a/odo/tests/test_append.py
+++ b/odo/tests/test_append.py
@@ -16,9 +16,9 @@ def test_append_list_to_set():
 
 def test_append_set_to_list():
     s = set([3, 4, 5])
-    l = [1, 2, 3]
-    append(l, s)
-    assert l == [1, 2, 3, 3, 4, 5]
+    lst = [1, 2, 3]
+    append(lst, s)
+    assert lst == [1, 2, 3, 3, 4, 5]
 
 
 def test_append_tuple_to_set():
--- a/odo/tests/test_utils.py
+++ b/odo/tests/test_utils.py
@@ -6,16 +6,18 @@ def test_ext():
     assert ext('apy') == ''
     assert ext('myfile.csv.gz') == 'gz'
 
+
 def test_iter_except():
-    s = set([1, 2, 3])   
+    s = set([1, 2, 3])
+
     def iterate():
         yield 1
         yield 2
         yield 4
-        
+
     elems = iterate()
     inset = iter_except(lambda: next(elems) in s, StopIteration)
-    assert list(inset) ==[True, True, False]
+    assert list(inset) == [True, True, False]
 
 
 def test_keywords():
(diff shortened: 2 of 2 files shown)


odo/tests/test_utils.py
```diff
@@ -1,7 +1,35 @@
...
+def test_gentemp():
+    i, _, data = next(gentemp([[1, 2], [3, 4]], start=1))
+    assert i==1 and data==[1, 2]
+
+
+def test_records_to_tuples():
+    seq = [{'i':1, 'f':1.0}, {'i':2, 'f':2.0}]
+    l = list(records_to_tuples('var * {i: int, f: float64}', seq))
+    assert l == [(1, 1.0), (2, 2.0)]
+    assert records_to_tuples('var * int', 'dummy') == 'dummy'
```
cpcloud: i didn't realize this was the behavior of `records_to_tuples` for things where the `dshape` doesn't match the type of the second argument. I'd probably expect an error to be raised. Can you put line 34 in a separate test, maybe `test_records_to_tuples_mismatched_input`?

cpcloud:
@ywang007 one minor comment, otherwise looks good!

odo/tests/test_append.py
```diff
@@ -12,3 +12,16 @@ def test_append_list_to_set():
     s = set([1, 2, 3])
     append(s, [4, 5, 6])
     assert s == set([1, 2, 3, 4, 5, 6])
+
+
+def test_append_set_to_list():
+    s = set([3, 4, 5])
+    l = [1, 2, 3]
```
llllllllll: minor bikeshedding comment, on some typefaces `1` and `l` are the same glyph, could we use names like `set_` and `list_` here? Note the suffixing underscore to not shadow the built in names.

odo/tests/test_utils.py
ywang007: I'll make a separate test. however per the comment in function, isn't this falls under the case of "non-record"?
>>> records_to_tuples('var * int', [1, 2, 3])  # pass through on non-records
    [1, 2, 3]

cpcloud: good point. i didn't actually look at the function, i was commenting as someone who didn't know exactly how it worked

odo/tests/test_append.py
cpcloud: i'd be okay with changing `l` -> `lst`

odo/tests/test_utils.py
```diff
@@ -1,7 +1,35 @@
...
+    assert list(inset) ==[True, True, False]
+
+
+def test_keywords():
+    assert keywords(keywords) == ["func"]
+
+
+def test_gentemp():
+    i, _, data = next(gentemp([[1, 2], [3, 4]], start=1))
+    assert i==1 and data==[1, 2]
```
llllllllll: can we put spaces around these binary operators?

odo/tests/test_utils.py
```diff
@@ -1,7 +1,35 @@
...
+def test_iter_except():
+    s = set([1, 2, 3])   
+    def iterate():
+        yield 1
+        yield 2
+        yield 4
+        
+    elems = iterate()
+    inset = iter_except(lambda: next(elems) in s, StopIteration)
+    assert list(inset) ==[True, True, False]
```
cpcloud: small nitpick: space after the `==` and before the `[`

odo/tests/test_utils.py
```diff
@@ -1,7 +1,35 @@
...
+    assert keywords(keywords) == ["func"]
+
+
+def test_gentemp():
+    i, _, data = next(gentemp([[1, 2], [3, 4]], start=1))
+    assert i==1 and data==[1, 2]
+
+
+def test_records_to_tuples():
+    seq = [{'i':1, 'f':1.0}, {'i':2, 'f':2.0}]
```
cpcloud: generally, we tend to put spaces after `:`

odo/tests/test_utils.py
ywang007: sure, i should have pep8 it. :)

odo/tests/test_utils.py
ywang007: will do.

cpcloud:
i find the the [autopep8](https://github.com/wistful/SublimeAutoPEP8) sublime text plugin useful as it lets me format single lines which is great for avoiding the urge to autopep8 the whole file.

cpcloud:
there's probably something similar for vim and emacs

llllllllll:
for emacs I use flycheck which will run flake8 over my python-mode buffers

ywang007:
uh, @cpcloud , i've made the changes per comments and pushed to my fork, a bit confused about what to do next, should i raise another PR? if so what should I do with this one? thanks.

cpcloud:
@ywang007 nope, no need to make another PR! just keep pushing to your fork, the commits show up as you push them

cpcloud:
@llllllllll any more comments before merge?

odo/tests/test_append.py
```diff
@@ -12,3 +12,16 @@ def test_append_list_to_set():
     s = set([1, 2, 3])
     append(s, [4, 5, 6])
     assert s == set([1, 2, 3, 4, 5, 6])
+
+
+def test_append_set_to_list():
+    s = set([3, 4, 5])
+    lst = [1, 2, 3]
+    append(lst, s)
+    assert lst == [1, 2, 3, 3, 4, 5]
```
cpcloud: do we know that this will always be

```python
[1, 2, 3, 3, 4, 5]
```
?

Could it ever end up being 

```python
[1, 2, 3, 4, 3, 5]
```
?

I think a test like this will be immune to this problem:

```python
s = set([3, 4, 5])
lst = [1, 2, 3]
newlist = lst[:]
append(lst, s)
newlist.extend(s)
assert lst == newlist
```

llllllllll:
No more comments from me, other than your comment on the set order it looks good

odo/tests/test_append.py
ywang007: Hmmm, for a set of int, since the hash() is just going to take int value, and we are not resizing the set here, so it seems the order will hold -- except when the implementation for set hash change?

so to be immune of that, how about:

s = set([3, 4, 5])
lst = [1, 2, 3]
append(lst, s)
assert sorted(lst) == [1, 2, 3, 3, 4, 5]

cpcloud: That's fine

ywang007: Done.

cpcloud:
@ywang007 looks good! merging

Issue #265, opened 2015-07-18
"Simulation mode" to show conversion paths in odo


octophat:
When writing new backends and debugging them, it would be useful if there was an option I could specify to the odo function that would "simulate" a conversion by printing out (or returning in a list) the function path it would take from source to destination without actually doing the conversion.

This could help discover issues without having to attempt to move data, and also shed light on potentially awkward paths odo may be taking to convert through the networkx graph to the target data type, allowing for more direct shortcuts to be taken. If nothing else, it could throw the curtain back on some of the magic going on behind the scenes and make the process more understandable.

I've wished this existed enough lately that I very well may just implement it, but welcome any thoughts on how it should be structured, how it should deal with append vs. convert (i.e. MultipleDispatch vs. networkx and cases where both are used), if the output should be printed/visualized/returned in a Python structure, etc.

mrocklin:
You might want to look at `odo.convert.path` (alias of `odo.core.path`, which holds the docstring) and `odo.append.resolve`, which together handle most of the internal path finding.  I'm not sure of the best way to present this information to the user but agree with you that this would be valuable.

octophat:
In the short term, I'm building this with print statements. In the long term, I don't think dumping a bunch of information via `print` is a good way for a widely-used functional package like this to behave, so I'm thinking of building a Simulation object that would be returned to the user from the odo function rather than the data object usually created, which would contain both an __str__ depiction that is useful if the user just wants printed information, as well as structured information about `append` and `convert` paths that would be useful for more automated testing or analysis about the paths odo takes.

As a first step though, I'm going with print statements just to make it work and understand what elements of the path are potentially useful and how they should be presented.

octophat:
First draft of simulation prints [here](https://github.com/octophat/odo/commit/bd814742a534179715ef9109e67ab0e38f810f08). Will try to organize that better in an object with a prettier print representation.

octophat:
I've gotten more experience with test-driven development lately that has me wanting to pick this up again. odo seems a bit difficult to use in a TDD framework because it does quite a bit of magic behind the scenes.

One typical testing pattern when developing a new backend would be to create a function that does a conversion or append, testing it in isolation outside of odo. Then the integration with odo can be tested by registering that function, calling odo, and verifying that the newly registered function was called with the expected arguments, using a mock for the new function so you can inspect it and not have it run.

[...]

PR #267, opened 2015-07-21
Simulation mode added to odo/into

octophat:
This change adds a simulate parameter to the into functions, which default to False, and are passed down implicitly from the odo function's kwargs. This should be useful when adding new backends, debugging problems, or optimizing conversion paths, to both enable path information to be returned in a usable format, and to provide a means for testing without having to actually initiate data conversions.

When simulate=True, instead of executing the functions returned by append and convert and returning the resulting data object, odo will create a new Path object, populate it with information about the conversion path(s) returned by append and convert, and return the Path.

This object can be printed for a human-readable string representation of the types, functions, and code that will be involved in the conversion, or a process can reference the information in the Path object's attributes.


octophat:
This addresses and resolves #265.

mrocklin:
Cool.  My first impression is that this adds a lot of code into what was previously a pretty small file/function that just dispatched off to more complex functions.  I wonder if there is a way to separate this simulation work out from `into.py` and into another file.  Similarly I wonder if we could pull out the code currently in the `into` functions to some separate `simulate` function, perhaps adding a small call to `into` to dispatch to that function as appropriate.  This helps by releasing future developers who aren't interested in simulate from having to sift through it to understand `into`.

octophat:
That's fair, let me see if I can split this off into a separate simulate.py.

On the Travis errors, I can fix the importlib error in 2.6, but the dask and bcolz errors seem unrelated to my changes... am I right?

octophat:
I think the last two commits keep `into.py` clean and get the simulation option working under Python 2 and 3. Let me know if there are any other changes to the functionality or style you would like.

octophat:
@cpcloud, I thought a little last night about having the simulation functionality implemented in a separate function (`odo.simulate()`) vs. as a parameter in the existing odo function (`odo.odo(simulate=True)`). I think that would require us to duplicate all the same multiple dispatch logic in into.py, to parse out the different types of inputs that the function could receive. It seems better to have it as a parameter to `odo.into` (and by extension, `odo.odo`) to piggyback off the existing logic. I think @mrocklin's concerns about cluttering up into.py should be OK now with the logic moved into simulate.py and the Path object.

I'm fine with changing the structure of the Path object or how it's represented, but it seems like invoking this through the `simulate=True` parameter causes the least duplication of code.

cpcloud:
@octophat taking a look at this now

octophat:
Apologies for not coming back to this sooner - I've been working on a different project unrelated to odo lately, but I have some time around the holidays now to come back to this a little wiser. :)

I'm going to withdraw this pull request for now as I want to take a fresh look at it, looking at the broader applicability of how odo can work in testing environments where you don't want to actually run conversions but need to verify that odo will take a specific path or run a given function. I think this is needed more than ever, but there'll be a better way to do it than the code I had here. I'll capture my thoughts in #265 and welcome any feedback.

PR #273, opened 2015-08-01
BUG: ndarray to df loses field names

llllllllll:
I couldn't run a transform on an InteractiveSymbol wrapping an ndarray because it pushed it into a df and lost all the column information

```python
In [1]: es = bz.Data(np.arange(15).reshape(5, 3), fields=list('abc'))

In [2]: bz.transform(es, d=es.a + es.b)
...
KeyError: 'a'
```


cpcloud:
Nice find. Merging

PR #275, opened 2015-08-05
BUG: infer_header was not forwarded the encoding

llllllllll:
https://github.com/wavexx/gtabview/issues/17


cpcloud:
Can you add a test for this?

llllllllll:
Sure thing

llllllllll:
I just made sure the `encoding.csv` file was `latin-1` and then added some non-utf code in there. This failed an existing test on master, we previously were using the intersection of utf-8 and latin-1 so it was fine.

cpcloud:
cool

cpcloud:
merge on pass

llllllllll:
Hmm, the test failure shows are not handing some sql encoding stuff correctly either. Looking into it.

llllllllll:
I am having trouble figuring out how to get sqlite and sqlalchemy to actually support latin1. Docs seems unclear.

cpcloud:
@llllllllll the travis failure is because you've written an encoded character in python code without telling python what encoding to use to read the file.

cpcloud:
write it as bytes and then call encode on it

llllllllll:
Ah, will be explicit about the encoding for py2 then.

cpcloud:
i think you can just do `b'the_chars'.decode('latin1')` and will work on all python versions

cpcloud:
@llllllllll can we salvage anything here? or should this be closed?

llllllllll:
The csv stuff in here was correct; however, I realized that there were some issues with encoding in sqlite. We could probably rip the the sql stuff out since the original issue I was going to fix was the csv encoding problems

PR #276, opened 2015-08-07
discover should not consume first element from chunks

llllllllll:


llllllllll:
How do I check what version of `toolz` this ran with, looks like it couldn't get the `peek` func

cpcloud:
why does the conda output look so horrible?

screen shot 2015-08-07 at 13 59 33

llllllllll:
no idea, does it not normally look like that?

cpcloud:
no it usually is newline delimited

llllllllll:
Any ideas on this, can't repro

llllllllll:
Looks like the newest version of toolz doesn't have 2.6 or 3.3 conda packages (https://github.com/pytoolz/toolz/issues/269)

mrocklin:
cc @ilanschnell .  Is there a reason why we would have built recent packages for `toolz` on 2.7 and 3.4 but not 2.6 and 3.3?

ilanschnell:
Yes.  Because we stopped supporting Python 2.6 and 3.3.  This means we are not building new (or updating old) Python packages ones against Python 2.6 and 3.3, unless requested by a customer.

mrocklin:
Hrm, can I request continued Python 2.6 and 3.3 builds for toolz?  The library is simple enough that it should hopefully not be an extra burden.  Alternatively would toolz be appropriate for a noarch build?

ilanschnell:
Why does odo still care about Python 2.6 and 3.3?

mrocklin:
Presumably people are still on 2.6.  I don't have a good view into this though.  

Note that `toolz` is depended upon far more widely than `odo`.  Even if `odo` doesn't add pressure for 2.6 there might be sufficient pressure from external projects.

PR #277, opened 2015-08-07
MAINT: adds copydoc decorator

llllllllll:
we have this pattern a lot which prevents us from using decorators as we would normally use them. I also ran into this issue with blaze where I couldn't mutate the `__doc__` of an unbound method in python 2. If this change looks good, I will refactor blaze to use this too.

### DIFF
--- a/docs/source/whatsnew/0.3.4.txt
+++ b/docs/source/whatsnew/0.3.4.txt
@@ -30,3 +30,12 @@ Bug Fixes
   a numpy array to a DataFrame (:issue:`273`).
 * Fix appending into a sql table from chunks not returning the table.
   (:issue:`278`).
+
+Miscellaneous
+-------------
+
+* Adds :func:`~odo.utils.copydoc` function to copy docstrings from one object
+  onto another. This helps with the pattern of explicitly setting the
+  ``__doc__`` attribute to the ``__doc__`` of another function or class. This
+  function can be used as a decorator like: ``@copydoc(FromThisClass)`` or as a
+  function like: ``copydoc(FromThisClass, to_this_function)``. (:issue:`277`).
--- a/odo/backends/aws.py
+++ b/odo/backends/aws.py
@@ -19,7 +19,7 @@ from multipledispatch import MDNotImplementedError
 from .text import TextFile
 
 from ..compatibility import urlparse
-from ..utils import tmpfile, ext, sample, filter_kwargs
+from ..utils import tmpfile, ext, sample, filter_kwargs, copydoc
 
 
 @memoize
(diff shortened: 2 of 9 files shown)


cpcloud:
this looks good, can you just confirm locally that you can see the docstring as desired?

cpcloud:
also pls add a blurb in whatsnew

llllllllll:
What section should I put this in?

llllllllll:
Looks like it is getting copied correctly
```python
In [10]: print(URL.__doc__)
 Parent class for data accessed through ``URLs``

    Parameters
    ----------

    url : str
        full url to file
    chunk_size : int (default 1024)
        Size of chunks streamed into memory
    decode_unicode : bool (default False)
        If True, content will be decoded using the best available
        encoding based on the response.

     Examples
    --------

    >>> from odo import URL, CSV
    >>> u = URL(CSV)('http://foo.com/myfile.csv')

    Normally create through resource uris

    >>> data = resource('http://foo.com/myfile.csv')
    >>> data.url
    'http://foo.com/myfile.csv'
    >>> data.filename
    'myfile.csv'
```

cpcloud:
> What section should I put this in?

Could make a new `Miscellaneous` section

docs/source/whatsnew/0.3.4.txt
```diff
@@ -28,3 +28,13 @@ Bug Fixes
 * Cope with Dask and bcolz API changes (:issue:`270`).
 * Fixed a bug where columns in dshape were being ignored when converting
   a numpy array to a DataFrame (:issue:`273`).
+
+
+Miscellaneous
+-------------
+
+* Adds ``~odo.utils.copydoc`` function to copy docstrings from one object onto
```
cpcloud: i think this needs to be ``:func:`~odo.utils.copydoc` ``

docs/source/whatsnew/0.3.4.txt
```diff
@@ -28,3 +28,13 @@ Bug Fixes
...
+  another. This helps with the pattern of explicitly setting the __doc__
```
cpcloud: the `__doc__`s need double backticks surrounding them

docs/source/whatsnew/0.3.4.txt
```diff
@@ -30,3 +30,12 @@ Bug Fixes
   a numpy array to a DataFrame (:issue:`273`).
 * Fix appending into a sql table from chunks not returning the table.
   (:issue:`278`).
+
+Miscellaneous
+-------------
+
+* Adds :func:``~odo.utils.copydoc`` function to copy docstrings from one object
```
cpcloud: i don't think double ticks will compile

cpcloud: sorry for the churn btw

llllllllll: I don't mind putting in the extra work for better docs. How do I build the docs locally to test the changes?

cpcloud: you need sphinx, then just `cd docs && make clean && make html`

cpcloud: then open `build/html/index.html` in your browser of choice

cpcloud:
LGTM

llllllllll:
I will open a PR soon on blaze that uses this new feature. That should fix that issue that we needed to revert with Node.map

cpcloud:
awesome thanks!

PR #280, opened 2015-08-13
Implement `raise_on_errors` kwarg for `drop`

cpcloud:
cc @jreback

### DIFF
--- a/docs/source/whatsnew/0.3.4.txt
+++ b/docs/source/whatsnew/0.3.4.txt
@@ -20,6 +20,11 @@ None
 API Changes
 -----------
 
+* The ``str`` dispatched version of the :func:`~odo.drop.drop` function now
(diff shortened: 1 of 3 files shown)


cpcloud:
@llllllllll can you give this a review when you get a chance?

llllllllll:
Implementation looks good, I just feel like the name is not very clear. People might get confused about which errors we are raising on. Maybe something like `ignore_nonexistent`.

kwmsmith:
@llllllllll this PR is pretty stale -- is it still relevant?

Regarding the argument name: other ways I've seen to approach this is to have an `errors` argument that is `'ignore'` by default, and can be set to `'raise'`, `'warn'`, etc, which provide more user control.  Thoughts?

llllllllll:
I like the raise, warn, ignore pattern; however, I prefer enums over strings. Harder to get wrong. Also, I think only raise or warn are sane defaults often

Issue #279, opened 2015-08-12
sqlite: NotImplementedError: Unable to parse uri to data resource:


earney:
I get 
NotImplementedError: Unable to parse uri to data resource: sqlite:///data/lahman2013.sqlite 

 until I install sqlalchemy.   I notice that requirements.txt does not contain sqlalchemy.  Should it?

llllllllll:
Odo understands a lot of different type conversions. In order to keep the dependencies small, many of them are optional and are only enabled when the particular backend is installed.This is an explicit decision. I think we could definitely do a better job documenting this though

earney:
After creating this issue, I discovered the optional_requirements.txt fle and saw sqlalchemy as a dependency.   
Would it be possible for the resource function to output a custom error message stating that sqlalchemy isn't installed?

earney:
maybe place something like:
````
    if uri.startswith('sqlite:'):
       _msg="Unable to parse uri to data resource: " + uri 
       _msg+="Do you have SqlAlchemy installed? (pip install SqlAlchemy)" 
       raise NotImplementedError(_msg)
```
at line 101, (just before the raise statement?)

Just an idea. :)

llllllllll:
The current system is to have all of the handling for a backend isolated in a module that registers handlers for `convert`, `resource`, and `discover`. odo was designed to be modular and extensible, and we take advantage of that internally by handling the default graph no differently than vertices that are added externally. In order to make these errors correct in the general case we would need to create a `resource` handler that dispatches on the same regex but just tells the use which import error was raised. I am not sure that the added complexity of doing all of this is worth it when it could be solved with documentation.

earney:
I think it is definitely worth seeing if adding some additional info to the docs would help.  I'm not seeing a place in the docs where I think this makes sense.  I've looked at sql.rst and uri.rst, and adding this type of information just seems to complicate the pages.

 What about a page in the docs for common issues or hiccups? If you think this is a good idea, I can try to put something together for you to inspect.
Billy

llllllllll:
Do you think that it would be helpful to have a section on the sql page of the docs that was about enabling these conversions? Something that mentioned that these will only be available if you have sqlalchemy installed?

If you wanted to work on a common issues page that would be awesome!

earney:
Ok. I'll start with adding a section about enabling the conversions on the sql page, and if that doesn't  seem right to us, we can look at a common issues page.

earney:
[...]

PR #281, opened 2015-08-13
enabling conversions

earney:
addresses issue #279,  Unable to parse uri to data source, even though uri is valid. This can occur when SQLAlchemy is not installed, so added additional documentation describing how to enable conversions.

### DIFF
--- a/docs/source/sql.rst
+++ b/docs/source/sql.rst
@@ -50,7 +50,7 @@ If SQLAlchemy is not installed and you try using one of the databases, you
 may get the following error:
 
 
-	NotImplementedError: Unable to parse uri to data resource: sqlite:///file.db
+    NotImplementedError: Unable to parse uri to data resource: sqlite:///file.db
 
 
 
--- a/docs/source/sql.rst
+++ b/docs/source/sql.rst
@@ -37,3 +37,23 @@ These are often an order of magnitude faster than the ``Python->SQLAlchemy``
 route when they are available.::
 
     sqlalchemy.Table <- CSV
+
+Enabling Conversions
+--------------------
+
+Odo is dependent on various conversion libraries.  Many of these are not 
+installed by default. 
+
+For example the SQLAlchemy library is used for sqlite, PostgreSQL, MySQL, Hive,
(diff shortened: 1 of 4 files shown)


docs/source/sql.rst
```diff
@@ -37,3 +37,16 @@ These are often an order of magnitude faster than the ``Python->SQLAlchemy``
...
+Enabling Conversions
+--------------------
+
+Odo is dependent on the SQLAlchemy library for sqlite, PostgreSQL, MySQL, Hive,
+and RedShift support.  SQLAlchemy is not installed by default when Odo is
+installed, therefore you must install it if you want to use one of these
+databases.
+
+If SQLAlchemy is not installed and you try using one of the databases, you
+may get the following error::
```
cpcloud: What do you think about making a general statement about enabling conversions, and then giving SQLAlchemy as a specific example.

I'd also like to include a shell command for installation, for completeness.

```sh
$ conda install sqlalchemy
$ pip install sqlalchemy
```

cpcloud: If you go the generalization route, you'd probably want to move this to a different file.

llllllllll:
I really like this paragraph; however, I feel like it should not be on the `sql` page. Maybe we could move this to the overview page? @cpcloud, thoughts?

cpcloud:
@llllllllll Yep, I wrong something along those lines in the outdate diff, so +1 on generalizing the language and possibly giving sqlalchemy as a specific example.

earney:
I've generalized and moved the paragraph to overview.rst.   Let me know if I need to make additional modifications.

llllllllll:
I think that you should keep the section about how to enable sql with the sqlalchemy example.

earney:
I agree!  I think it doesn't hurt to repeat some of the info, especially if we get into specifics (like we do for SQLAlchemy).
@cpcloud , do you agree?

cpcloud:
+1

earney:
Okay. I'll push a fix.

llllllllll:
pulled the docs changes in with some tweaks with 02a77bddfe8627e0bfe342bd05b8988be9c0707b

PR #282, opened 2015-08-14
ENH: glob for S3 CSVs

TomAugspurger:
@cpcloud this would be super useful for me. Redshift dumps a (gzipped) CSV per node.

My questions:

- priority? I think I took the same as the CSV glob
- Better way to list the keys matching the glob than creating a new S3 connection inside `resource_s3_csv_glob`?
- ~~It doesn't actually work yet... is there another step to hook this up to the network?~~ 
- tests
- is there a better path to hook into than directly from `chunks(s3(CSV)) -> chunks(DataFrame)`?

I've only spent 1-2 hours on this and I'll continue to work on it. I'm just guessing that some of these will be easier for you to answer :)

### DIFF
--- a/odo/backends/aws.py
+++ b/odo/backends/aws.py
@@ -4,13 +4,13 @@ import os
 import uuid
 import zlib
 import re
-
+from fnmatch import fnmatch
 from contextlib import contextmanager
 from collections import Iterator
 from operator import attrgetter
 
 import pandas as pd
-from toolz import memoize, first
+from toolz import memoize, first, concat
 
 from .. import (discover, CSV, resource, append, convert, drop, Temp, JSON,
                 JSONLines, chunks)
@@ -151,6 +151,28 @@ def resource_s3_csv(uri, **kwargs):
     return S3(CSV)(uri, **kwargs)
 
 
+@resource.register('s3://.*\*.csv(\.gz)?', priority=19)
+def resource_s3_csv_glob(uri, **kwargs):
+    con = get_s3_connection()
+    result = urlparse(uri)
+    bucket = con.get_bucket(result.netloc)
+    key = result.path.lstrip('/')
+
+    all_keys = bucket.list(prefix=key.split('*')[0])
+    matched = [k for k in all_keys if fnmatch(k.key, key)]
+    uris = ['s3://{}/{}'.format(bucket.name, k.key) for k in matched]
+    r = resource(uris[0], **kwargs)
+    return chunks(type(r))([resource(u, **kwargs) for u in sorted(uris)])
+
+
+@convert.register(chunks(pd.DataFrame), chunks(S3(CSV)), cost=11.0)
+def convert_gob_of_s3_csvs_to_chunks_of_dataframes(csvs, **kwargs):
+   def _():
+       return concat(convert(chunks(pd.DataFrame), csv, **kwargs)
+                      for csv in csvs)
+   return chunks(pd.DataFrame)(_)
+
+
 @resource.register('s3://.*\.txt(\.gz)?', priority=18)
 def resource_s3_text(uri, **kwargs):
     return S3(TextFile)(uri)
@@ -272,3 +294,5 @@ else:
             raise ValueError('Error code %d, message: %r' % (exit_status,
                                                              stderr.read()))
         return ssh
+
+
--- a/docs/source/aws.rst
+++ b/docs/source/aws.rst
@@ -33,12 +33,23 @@ Interface
 URIs
 ----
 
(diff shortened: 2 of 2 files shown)


TomAugspurger:
Ah, I see why it wasn't getting added to the graph. `convert` expects (destination, source), not (source, destination). A leftover from `into` I'm guessing.

mrocklin:
Indeed

odo/backends/aws.py
```diff
@@ -138,6 +138,26 @@ def resource_s3_csv(uri, **kwargs):
     return S3(CSV)(uri, **kwargs)
 
 
+@resource.register('s3://.*\*.csv(\.gz)?', priority=19)
+def resource_s3_csv_glob(uri, **kwargs):
+    import boto
```
cpcloud: You can put this at the top level

cpcloud:
@TomAugspurger what do you think about generalizing a bit here and making a `Bucket` type?

cpcloud:
maybe something like

```python
from odo import Bucket
bucket = resource('s3://path/to/my/bucket/*.csv')  # type(bucket) == Bucket(S3(CSV))
bucket.name == 'bucket'
bucket.path == 'path/to/my/my/bucket'
```

cpcloud:
> Better way to list the keys matching the glob than creating a new S3 connection inside `resource_s3_csv_glob`?

Call `get_s3_connection` which memoizes the connection keyed by the requisite AWS credentials

cpcloud:
> priority? I think I took the same as the CSV glob

yours should be higher priority, since you want yours to match before matching a plain CSV file

cpcloud:
Hm, I wonder if we could make this work with `boto.Bucket` objects...

cpcloud:
Ok, scratch the `boto.Bucket` idea for now ... doesn't look like there's a way to make the bucket look as if it has a particular set of files, which we need to do if we wanted to use bucket + globbing.

@TomAugspurger I'm think the chunks approach is nice.

TomAugspurger:
Ok, refactored a bit using your comments. I'm reusing the s3 connection, added / tweaked the S3 docs a bit. Tests are still lacking :/ but it works for now at least.


```python
In [1]: import odo

In [2]: odo.odo('s3://vehcuity/tst*.csv.gz', 'test.csv')
Out[2]: <odo.backends.csv.CSV at 0x105a58358>

In [3]: !cat test.csv
0,1
0.07331551328957626,-0.4154916469412773
-1.562069894747674,-1.4722870727590358
-0.15131201641683129,-0.31657923775839325
-0.666743351952952,-1.0564613439645762

In [4]: r = odo.resource('s3://vehcuity/tst*.csv.gz')

In [5]: r
Out[5]: <odo.chunks.chunks(odo.S3(CSV)) at 0x105ae4748>

In [6]: r.
r.container  r.data

In [6]: r.data
Out[6]:
[<odo.backends.aws.S3(CSV) at 0x105ae4390>,
 <odo.backends.aws.S3(CSV) at 0x105ae42b0>]
````

File: odo/backends/aws.py [hunk 2]
```diff
@@ -138,6 +138,37 @@ def resource_s3_csv(uri, **kwargs):
...
+@resource.register('s3://.*\*.csv(\.gz)?', priority=19)
+def resource_s3_csv_glob(uri, **kwargs):
+    con = get_s3_connection()
+    result = urlparse(uri)
+    bucket = con.get_bucket(result.netloc)
+    key = result.path.lstrip('/')
+
+    all_keys = bucket.list(prefix=key.split('*')[0])
+    matched = _filter_to_matched(key, all_keys)
+    uris = [_key_to_uri(k, bucket) for k in matched]
```
cpcloud: i think it's okay to inline `_filter_to_matched` and `_key_to_uri`, IMO it's pretty clear what they are doing

odo/backends/aws.py
```diff
@@ -138,6 +138,37 @@ def resource_s3_csv(uri, **kwargs):
...
 
 
+@resource.register('s3://.*\*.csv(\.gz)?', priority=19)
+def resource_s3_csv_glob(uri, **kwargs):
+    con = get_s3_connection()
+    result = urlparse(uri)
+    bucket = con.get_bucket(result.netloc)
+    key = result.path.lstrip('/')
+
+    all_keys = bucket.list(prefix=key.split('*')[0])
```
cpcloud: awesome! i needed this for https://github.com/blaze/odo/pull/293/files

TomAugspurger:
@cpcloud inlined those to functions, I had them broken out for testing, since they don't do any IO, but then gave up on testing :) I need to learn how mock works.

cpcloud:
Fixing this up and merging. I've added a test

PR #284, opened 2015-08-17
Add postgres ARRAY type discovery and creation ability

cpcloud:
closes https://github.com/ContinuumIO/blaze/issues/1210

This significantly uglifies dshape_to_alchemy (which was already getting gross).
It might be time for that to be refactored before the next release

### DIFF
--- a/odo/backends/sql.py
+++ b/odo/backends/sql.py
@@ -142,6 +142,11 @@ def discover_postgresql_interval(t):
                                       second_precision=t.precision))
 
 
+@discover.register(sa.dialects.postgresql.base.ARRAY)
+def discover_postgresql_array(t):
+    return var * discover(t.item_type)
+
+
 @discover.register(sa.dialects.oracle.base.INTERVAL)
 def discover_oracle_interval(t):
     return discover(t.adapt(sa.types.Interval))
@@ -269,7 +274,7 @@ def dshape_to_table(name, ds, metadata=None):
         ds = dshape(ds)
     if metadata is None:
         metadata = sa.MetaData()
-    cols = dshape_to_alchemy(ds)
+    cols = dshape_to_alchemy(ds, getattr(metadata.bind, 'dialect', None))
     t = sa.Table(name, metadata, *cols, schema=metadata.schema)
     return attach_schema(t, t.schema)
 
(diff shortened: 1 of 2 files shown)


rabernat:
:+1:

odo/backends/tests/test_postgres.py
```diff
@@ -91,8 +92,24 @@ def complex_sql(url):
...
+            yield t
+        finally:
+            drop(t)
+
+
+@pytest.yield_fixture
+def sql_with_array(url):
+    ds = 'var * {a: var * float64, b: string}'
+    try:
+        t = resource(url, dshape=ds)
```
llllllllll: do you think we should explicitly create the sql table here. If there is a bug where this doesn't actually create an array but is also not discovered as an array then the test will pass.

cpcloud: yes that's a good idea, and i'll add a separate test for creating the table with odo

odo/backends/sql.py
```diff
@@ -307,19 +317,34 @@ def dshape_to_alchemy(dshape):
...
+                              dshape_to_alchemy(typ, dialect, level=level + 1),
+                              nullable=isinstance(typ, Option))
+                    for name, typ in dshape.parameters[0]]
+        raise TypeError(
+            'nested record dshape not supported by sqlalchemy')
     if isinstance(dshape, datashape.DataShape):
         if isdimension(dshape[0]):
-            return dshape_to_alchemy(dshape[1])
+            if isscalar(dshape[1]):
+                type_ = array_types.get(dialect.name, None)
```
llllllllll: `None` is the default arg here

cpcloud: Yeah I'll move it up so that the `dialect` appears as `dialect.name`

odo/backends/sql.py
```diff
@@ -307,19 +317,34 @@ def dshape_to_alchemy(dshape):
...
+                if type_ is None:
+                    raise TypeError('dialect %s has no array type for dshape '
+                                    '%s' % (dialect.name, dshape))
+                return type_(dshape_to_alchemy(dshape[1],
+                                               dialect, level=level + 1))
+            else:
+                if not level:
+                    return dshape_to_alchemy(dshape[1], dialect,
+                                             level=level + 1)
+                raise TypeError('nested record dshape not supported by sqlalchemy')
```
llllllllll: Is this going to make it more difficult to implement your foreign key feature?

cpcloud: Yes, thanks for pointing that out. I'll merge this into that and see what happens.

llllllllll:
If someone wants to own this please reopen

PR #286, opened 2015-08-18
doc and error for existing non-appendable target

### LINKED ISSUE #283: can not use an existing DataFrame as target in odo(source, target)
It seems an existing DataFrame can't be used as target, even when its datashape matches source, not sure if this is intentional, or a bug?

for example:
```
csv = resource('test.csv')
df = odo(csv, pd.DataFrame) #initialize same as csv
odo(df, csv) #fine, simply append df content to csv
odo(csv, df) #throws NotImplementedError 
```
among csv, list, db (sqlite), they all can convert (append) between each other without issue, but none of them can be converted (append) to existing DataFrame. 
Specifically:
self appending -- odo(x,x) works fine for csv, list, sqlite etc, but
odo(df, df) does not work.

ywang007:
close #283 

I added the ValueError and doc for existing target that does not support in-place append. Please check.

Is there a better way than enumerate DataFrame, ndarray etc?

### DIFF
--- a/docs/source/overview.rst
+++ b/docs/source/overview.rst
@@ -43,6 +43,12 @@ So the following lines would be valid inputs to ``odo``
    >>> odo('postgresql://hostname::tablename', 'myfile.json') # Dump Postgres to JSON
    >>> odo('mongodb://hostname/db::collection', pd.DataFrame) # Dump Mongo to DataFrame
 
+Please note that, if target already exists, it must be of a type that 
+supports in-place append.
+
+.. code-block:: python
+
+   >>> odo('myfile.csv', df) # this will raise ValueError because DataFrame is not appendable
 
 Network Effects
 ---------------
--- a/docs/source/overview.rst
+++ b/docs/source/overview.rst
@@ -48,7 +48,7 @@ supports in-place append.
 
 .. code-block:: python
 
-   >>> odo('myfile.csv', df) # this will raise ValueError because DataFrame is not appendable
+   >>> odo('myfile.csv', df) # this will raise TypeError because DataFrame is not appendable
 
 Network Effects
 ---------------
--- a/odo/into.py
+++ b/odo/into.py
@@ -98,7 +98,7 @@ def into_object(target, source, dshape=None, **kwargs):
     if isinstance(source, (str, unicode)):
         source = resource(source, dshape=dshape, **kwargs)
     if type(target) in not_appendable_types:
-        raise ValueError('target of %s type does not support in-place append' % type(target))
+        raise TypeError('target of %s type does not support in-place append' % type(target))
     with ignoring(NotImplementedError):
         if dshape is None:
             dshape = discover(source)
--- a/odo/into.py
+++ b/odo/into.py
@@ -10,6 +10,8 @@ from datashape import discover
 from datashape.dispatch import namespace
 from datashape.predicates import isdimension
 from .compatibility import unicode
+from pandas import DataFrame, Series
+from numpy import ndarray
 
 
 if 'into' not in namespace:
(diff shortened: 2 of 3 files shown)


odo/into.py
```diff
@@ -94,6 +96,8 @@ def into_object(target, source, dshape=None, **kwargs):
     """
     if isinstance(source, (str, unicode)):
         source = resource(source, dshape=dshape, **kwargs)
+    if isinstance(target, (DataFrame, Series, ndarray)):
```
cpcloud: suggest a global variable like this:

```python
not_appendable = DataFrame, Series, ndarray, tuple

...
    if isinstance(target, not_appendable):
        raise TypeError('target of type %r does not support in-place append' % type(target).__name__)
```

llllllllll:
closing and reopening to see the test output. looks like it is pointing to the old name,

ywang007:
@llllllllll which "old name"?   Some tests which ran OK before suddenly fail now, was it some environment/software version change? Do you have a solution for that? Thanks.

llllllllll:
sorry, the link for the test was incorrect based on the organisation change from continuumio to blaze.

The tests are failing because `EmptyHDFStoreDataset` is a subclass of tuple.

```python
EmptyHDFStoreDataset = namedtuple('EmptyHDFStoreDataset', 'parent,pathname,dshape')
```
I am wondering if the cost of maintaining this set of unappendables is worth the slightly better error message. Subtle things like this could crop up in user defined types. What if a user is dispatching on their own subclasses of these types?

ywang007:
Yes, checking "type(target) in not_appendable_types” seems better. Even though it let pass those subclass which is indeed still not-appendable, at least it won’t raise false alarm and fail when the target is good.

odo/tests/test_into.py
```diff
@@ -20,6 +20,11 @@ def test_into_append():
     assert result is lst
 
 
+def test_into_append_failure():
+    nd = np.array([1, 2, 3])
+    assert raises(ValueError, lambda: odo((4, 5), nd))
```
cpcloud: can you write this as

```python
```

odo/into.py
```diff
@@ -94,6 +97,8 @@ def into_object(target, source, dshape=None, **kwargs):
     """
     if isinstance(source, (str, unicode)):
         source = resource(source, dshape=dshape, **kwargs)
+    if type(target) in not_appendable_types:
+        raise ValueError('target of %s type does not support in-place append' % type(target))
```
cpcloud: I think this should be a `TypeError`

docs/source/overview.rst
```diff
@@ -43,6 +43,12 @@ So the following lines would be valid inputs to ``odo``
    >>> odo('postgresql://hostname::tablename', 'myfile.json') # Dump Postgres to JSON
    >>> odo('mongodb://hostname/db::collection', pd.DataFrame) # Dump Mongo to DataFrame
 
+Please note that, if target already exists, it must be of a type that 
+supports in-place append.
+
+.. code-block:: python
+
+   >>> odo('myfile.csv', df) # this will raise ValueError because DataFrame is not appendable
```
cpcloud: `TypeError`

cpcloud:
@ywang007 couple of small comments, other than those good 2 go!

ywang007:
@cpcloud changed per your comments. Thanks.

odo/tests/test_into.py
```diff
@@ -20,6 +20,12 @@ def test_into_append():
     assert result is lst
 
 
+def test_into_append_failure():
+    with pytest.raises(TypeError):
+        nd = np.array([1, 2, 3])
```
cpcloud: keep this outside of the `with` statement, as we only want to catch errors triggered by the `odo` call

odo/tests/test_into.py
```diff
@@ -48,3 +54,5 @@ def test_into_string_dshape():
     np.testing.assert_array_equal(odo([1, 2, 3], np.ndarray,
                                       dshape='var * float64'),
                                   np.array([1, 2, 3], dtype='float64'))
+
+test_into_append_failure()
```
cpcloud: don't call this explicitly, pytest does that for you

cpcloud:
@ywang007 thanks! this is in

PR #299, opened 2015-08-27
Make unit_to_dtype aware of Option(Decimal) types

octophat:
The Decimal type I introduced in blaze/datashape#170 has `to_numpy_dtype()` defined to downcast the Decimal types to NumPy ints and floats, but odo will not recognize Option versions of these Decimals without this fix. I also added tests to verify this.

This does introduce a dependency on the datashape version that will contain Decimals. Not sure if that needs to be addressed here or not... I was thinking of wrapping the `unit_to_dtype` new code in a `with ignoring(NameError):` block, which will solve this issue, but the tests will still fail in that case.

### DIFF
--- a/odo/numpy_dtype.py
+++ b/odo/numpy_dtype.py
@@ -23,6 +23,8 @@ def unit_to_dtype(ds):
     if isinstance(ds, DataShape):
         ds = ds.measure
     if isinstance(ds, Option) and isscalar(ds) and isnumeric(ds):
+        if isinstance(ds.ty, Decimal):
+            return unit_to_dtype(str(ds.ty.to_numpy_dtype()).replace('int', 'float'))
         return unit_to_dtype(str(ds).replace('int', 'float').replace('?', ''))
     if isinstance(ds, Option) and isinstance(ds.ty, (type(date_),
             type(datetime_), type(string), type(timedelta_))):
--- a/odo/numpy_dtype.py
+++ b/odo/numpy_dtype.py
@@ -4,6 +4,8 @@ import numpy as np
 from datashape import *
 from datashape.predicates import isscalar, isnumeric
 
+from .utils import ignoring
+
 def unit_to_dtype(ds):
     """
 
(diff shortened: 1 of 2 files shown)


cpcloud:
Hm, why is this change necessary?

octophat:
My understanding is that NumPy only accepts missing float values and not missing int values, hence the existing line changing ints to floats and removing the "?" for other numeric values. This won't work as coded for Decimal because it will be converted to either an int or a float only upon calling `to_numpy_dtype()`. Nullable decimals should still be supported though, we just need to downcast them to ints or floats and then use the same logic used to make ints and floats missing-compatible.

Does that make sense?

octophat:
@cpcloud, with blaze/datashape#170 now merged, this PR passes in Travis and should be merged. This will allow `Option(Decimal)` to work in DataFrames by extending the int -> float conversion for nullable/missing fields.

odo/numpy_dtype.py
```diff
@@ -23,6 +25,9 @@ def unit_to_dtype(ds):
     if isinstance(ds, DataShape):
         ds = ds.measure
     if isinstance(ds, Option) and isscalar(ds) and isnumeric(ds):
+        with ignoring(NameError):
+            if isinstance(ds.ty, Decimal):
+                return unit_to_dtype(str(ds.ty.to_numpy_dtype()).replace('int', 'float'))
```
cpcloud: why are the above 3 changes necessary? shouldn't this just work?

octophat: I'll try to confirm with a concrete example later today or tomorrow, but the existing code stringifies the subtype of the Option and then replaces int with float, but if the type is `Option(Decimal)`, this code won't apply, it won't strip off the Option type, and will end up triggering the Option `to_numpy_dtype` error.

I'll try to verify this when I'm not on my phone. :) I believe I encountered this in testing after adding `Decimal`.

octophat: OK, it doesn't return an error as-is. What `unit_to_dtype` does for an Option(Decimal(9, 0)) is this:

1. `if isinstance(ds, Option) and isscalar(ds) and isnumeric(ds):` will be True
2. It evaluates `str(ds).replace('int', 'float').replace('?', '')`, which gives `decimal[9, 0]` (it strips off the '?')
3. It calls `unit_to_dtype('decimal[9, 0]')`, which returns `dtype('int32')`.

What we'd expect, since this is an Option int, is for it to return a `dtype('float32')`. This doesn't happen because the replacement of `int` with `float` doesn't take place since the Decimal isn't revealed as an `int` until after the replace calls. So this change does the replace on `str(ds.ty.to_numpy_dtype())` instead for Option(Decimal) types. This preserves the behavior that Option ints will be floats in pandas Dataframes.

cpcloud: @octophat , cool can you remove the `ignoring(NameError)` block and we can get this in?

cpcloud: or i can make this change and merge if you don't have any cycles to spare

cpcloud:
@octophat thanks!

PR #301, opened 2015-08-28
Only return MetaData from resource

cpcloud:


cpcloud:
closing as low priority right now

PR #302, opened 2015-08-28
Make all strings in the sql backend unicode types unless they are explicitly ascii

cpcloud:


kwmsmith:
@llllllllll thoughts on this one?  You have the relevant experience here...

PR #310, opened 2015-09-05
fixes #309 - replace last utf-8 character when infering header

### LINKED ISSUE #309: csv header sniffer fails for utf-8 encoding
Because the method to infer the header in the file takes exactly x bytes it can fail if the last byte is part of a multi-byte utf-8 character. The decoding of the first x bytes should ignore the errors see the [codecs](https://docs.python.org/3.4/library/codecs.html#error-handlers) documentation.

A possible solution is to replace invalid bytes with the unicode replacement character:

~~~~{diff}
--- a/odo/backends/csv.py
+++ b/odo/backends/csv.py
@@ -93,7 +93,7 @@ def infer_header(path, nbytes=10000, encoding='utf-8', **kwargs):
     if not raw:
         return True
     sniffer = PipeSniffer()
-    decoded = raw if PY2 else raw.decode(encoding)
+    decoded = raw if PY2 else raw.decode(encoding, 'replace')
     try:
         return sniffer.has_header(decoded)
     except csv.Error:

~~~~

hdevalke:
closes #309

### DIFF
--- a/odo/backends/csv.py
+++ b/odo/backends/csv.py
@@ -93,7 +93,7 @@ def infer_header(path, nbytes=10000, encoding='utf-8', **kwargs):
     if not raw:
         return True
     sniffer = PipeSniffer()
-    decoded = raw if PY2 else raw.decode(encoding)
+    decoded = raw if PY2 else raw.decode(encoding, 'replace')
     try:
         return sniffer.has_header(decoded)
     except csv.Error:


cpcloud:
Can you add a test that fails on master and passes on this PR? Thanks!

cpcloud:
Should this also go here?

https://github.com/hdevalke/odo/blob/issue/309/odo/backends/csv.py#L113

cpcloud:
I cherry picked your commit into #311 ... closing this PR.

PR #311, opened 2015-09-05
Fix multibyte encoding issue

### LINKED ISSUE #309: csv header sniffer fails for utf-8 encoding

cpcloud:
closes #309

### DIFF
--- a/docs/source/whatsnew/0.3.4.txt
+++ b/docs/source/whatsnew/0.3.4.txt
@@ -52,6 +52,8 @@ Bug Fixes
 * Fix a bug where the SparkSQL backend wouldn't work if we didn't have paramiko installed (:issue:`300`)
 * Fix a testing bug where the endlines were being compared and they
   shouldn't have been (:issue:`312`).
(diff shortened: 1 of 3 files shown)


cpcloud:
merging on pass

PR #316, opened 2015-09-10
Quoted MySQL table name when using LOAD DATA INFILE

alexmojaki:


cpcloud:
@alexmojaki can you add a test for this change?

alexmojaki:
@cpcloud I can't see any existing tests for loading data into SQL (although I don't know this repo well so it may be right there) and I'm not going to create one from scratch. If you want to build on an existing test then try a table name with a non-standard character such as `-` in it (this is how I ran into the problem).

cpcloud:
Closing in favor of #371

PR #322, opened 2015-09-15
ENH: pass convert function the edges they are connecting

llllllllll:
I am trying to use some proxy types in a graph.

Ideally I could create a class like:

``` python
class Any(metaclass=ABCMeta):
    @classmethod
    def __subclasshook__(cls, subcls):
        return True
```

and then write a some converter like:

``` python
@convert.register(Any, Proxy, cost=1.0)
def _(p, converting_to=None, **kwargs):
    if isinstance(p.data, converting_to):
        return p.data
    raise NotImplemented()
```

to hook my proxy into the graph.

As a heuristic for now, I can replace `Any` with `object.__subclasses__()` (though that is sort of scary

Also, if you have a suggestion for a better way to do this, please let me know.

### DIFF
--- a/odo/core.py
+++ b/odo/core.py
@@ -38,12 +38,19 @@ def _transform(graph, target, source, excluded_edges=None, ooc_types=ooc_types,
     with ignoring(NotImplementedError):
         if 'dshape' not in kwargs:
             kwargs['dshape'] = discover(x)
+
     pth = path(graph, type(source), target,
                excluded_edges=excluded_edges,
                ooc_types=ooc_types)
     try:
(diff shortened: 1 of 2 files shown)


cpcloud:
Should this be done by changing the signature of the function that convert is decorating? So, instead of:

```python
@convert.register(A, B)
def converter(b, **kwargs):
    pass
```

We'd have

```python
@convert.register(A, B)
def converter(a, b, **kwargs):
    pass
```

similar to the signature of `append`.

llllllllll:
idk if I will ever get back to this, closing to clean up the number of stale prs.

PR #323, opened 2015-09-16
ENH: Use a warning instead of print

llllllllll:
I would prefer to be able to control where this message goes


cpcloud:
Small test and this is good to go. Rls note 2.

llllllllll:
merging on pass

PR #330, opened 2015-10-02
ENH: Adds castra support

llllllllll:
ping @jcrist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,6 +61,7 @@ install:
   - conda uninstall toolz
   - pip install -U toolz sas7bdat psycopg2 dill 'pymongo<3'
   - pip install git+git://github.com/blaze/dask.git --upgrade
+  - pip install git+git://github.com/blaze/castra.git --upgrade
   - if [ -n "$PANDAS_VERSION" ]; then pip install $PANDAS_VERSION; fi
 
   # install pyspark
--- a/docs/source/whatsnew/0.4.0.txt
+++ b/docs/source/whatsnew/0.4.0.txt
@@ -22,7 +22,6 @@ Experimental Features
   currently stable. This feature may break at any time without warning.
   (:issue:`330`).
 
-None
 
 New Backends
 ------------
--- a/docs/source/whatsnew/0.4.0.txt
+++ b/docs/source/whatsnew/0.4.0.txt
@@ -16,6 +16,12 @@ Experimental Features
 
    Experimental features are subject to change.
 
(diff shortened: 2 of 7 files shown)


jcrist:
I'm not super familiar with odo, but on first glance this looks good from the castra side of things.

mrocklin:
As a warning Castra is an experimental codebase and the API may change significantly and without warning.

llllllllll:
I understand the state of castra right now. I wanted to try prototyping using dask with a system that already uses odo so I figured I could push this back into odo.
Maybe we could have these functions raise `ExperimentalWarning` or something that alerts the user that this might not work with their version of castra.

llllllllll:
ping @cpcloud , is this okay to merge?

cpcloud:
Please add this feature to the experimental section with a note saying that castra itself is subject to change formats without warning.

cpcloud:
Also a couple of follow ups to think about but not necessary for this PR:

1. Categorical dshapes which are on the way. Castra becomes much more useful when categoricals can be used.
2. Adding `castra://` to the resource dispatcher.

cpcloud:
Maybe add an X failing test for odoing a castra to a dataframe when the castra has categories.

cpcloud:
Then good to merge

cpcloud:
@llllllllll anything left to do here?

llllllllll:
I don't think I have gotten to the feedback you left about the whatsnew or tests. I will implement that before merging unless you need this now.

llllllllll:
Also, I should test on the newest version of castra. I haven't touched this in a while.

cpcloud:
cool, no rush

llllllllll:
Quick question about the resource dispatcher, should we have `castra://` and `.castra`? Currently I only have the `.castra`

cpcloud:
probably should have both since presumably not every `Castra`'s file name will end in `.castra`

llllllllll:
I forgot about this. Just pushed the updates and rebased but looks like dask tests are failing ;_;

cowlicks:
@llllllllll I'll look at the dask tests today. The failures are not a result of your PR.

kwmsmith:
@llllllllll I'm holding off on adding new backends for the 0.4.1 release -- when you get a chance, could you look into the Travis-CI failure?

llllllllll:
Sure thing. I also need to rebase

llllllllll:
I have been told to not use castra anymore so it is probably not worth merging this.

PR #331, opened 2015-10-05
ENH: Adds nan->pd.NaT edge

llllllllll:
Currently reductions on series of dtype datetime64 which are empty result in a `nan` instead of ` pd.NaT` which causes things like:

```python
odo(pd.Series(dtype='datetime64[ns]').max(), pd.Timestamp)
```

to fail by not finding a route from `float -> pd.Timestamp`

I have a pr for pandas (https://github.com/pydata/pandas/pull/11245) that changes return to nat in pandas land but this solves the immediate issue.

### DIFF
--- a/odo/backends/tests/test_pandas.py
+++ b/odo/backends/tests/test_pandas.py
@@ -4,7 +4,9 @@ from datetime import datetime
 
 from datashape import discover, Option
 from datashape import dshape
+from networkx import NetworkXNoPath
 import pandas as pd
+import pytest
 
 from odo import odo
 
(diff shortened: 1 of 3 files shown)


odo/backends/tests/test_pandas.py
```diff
@@ -37,3 +39,11 @@ def test_datetime_to_timestamp():
     ts = odo(dt, pd.Timestamp)
     assert isinstance(ts, pd.Timestamp)
     assert ts == pd.Timestamp('2014-01-01')
+
+
+def test_nan_to_nat():
+    assert odo(float('nan'), pd.Timestamp) is pd.NaT
+
```
cpcloud: Can you add a test that this works with numpy nans as well?

cpcloud: Or just a single line to this test is fine.

llllllllll: kk, though np.nan is just `PyFloat_FromDouble((float) 0x7fc00000UL)` so it is not a custom object.

PR #338, opened 2015-10-15
Fix decimal discovery for MySQL

cpcloud:


cpcloud:
@llllllllll any comments here? otherwise merging on pass

llllllllll:
looks good. I don't have a local mysql instance but this seems correct.

Issue #340, opened 2015-10-23
Pathological runtimes with specific CSV files


kwmsmith:
As reported by @jbednar 

The file `data.csv` has 3 named columns with trailing commas and a trailing space.  `pandas.read_csv()` can handle this fine, while `odo` completes, but takes several hundred times longer to complete.

```
$ head data.csv
a,b,c, 
-1.3700737E7, 6275190.0, w, 
-1.3700711E7, 6275195.0, w, 
-1.3702081E7, 6274898.5, w, 
-1.3701948E7, 6274931.0, w, 
-1.3701793E7, 6275088.5, w, 
-1.3701224E7, 6274994.5, w, 
-1.370176E7, 6274780.0, w, 
-1.370215E7, 6274539.5, w, 
-1.3702123E7, 6274279.0, w, 

$ wc -l data.csv
  306675005 data.csv
```

`odo`ing the first 10**6 lines of this file into a `DataFrame` takes approximately 30s, meaning the entire 300M line file will take upwards of 2.5 hours to load.

```python
In [4]: %time odo('data_1E6.csv', DataFrame)
CPU times: user 29.7 s, sys: 74.6 ms, total: 29.8 s
Wall time: 29.8 s
Out[4]:
        a            b             c Unnamed: 3
0        -13700737    6275190.0    w 2015-10-23
1        -13700711    6275195.0    w 2015-10-23
2        -13702081    6274898.5    w 2015-10-23
3        -13701948    6274931.0    w 2015-10-23
4        -13701793    6275088.5    w 2015-10-23
5        -13701224    6274994.5    w 2015-10-23
6        -13701760    6274780.0    w 2015-10-23
7        -13702150    6274539.5    w 2015-10-23
8        -13702123    6274279.0    w 2015-10-23
9        -13702072    6274537.0    w 2015-10-23
10       -13701917    6274562.5    w 2015-10-23
```

The equivalent `pandas.read_csv()` call takes approximately 3 minutes to load *the entire file*.

The `odo` call has trouble with the trailing commas and trailing spaces.  It extracts an unnamed column, and the key performance killer is that odo [*converts the space in that column into a datetime*](https://github.com/blaze/odo/blob/master/odo/backends/pandas.py#L60-L61).

Fix is forthcoming.

jbednar:
Ouch!  Thanks for figuring that out; it will be a big help!

cpcloud:
@kwmsmith @jbednar I'm unable to reproduce this performance issue.

[...]

PR #342, opened 2015-10-23
Fix whitespace column datetime coercion for pandas < 0.17

kwmsmith:
This patches `pandas.to_datetime()` for pandas < 0.17.  Does not affect things for Pandas == 0.17.

Resolves issue #340.

### DIFF
--- a/odo/backends/pandas.py
+++ b/odo/backends/pandas.py
@@ -58,18 +58,19 @@ def coerce_datetimes(df):
     dtype: object
     """
 
-    df2 = df.select_dtypes(include=[object])
-    # NOTE: In pandas < 0.17, pd.to_datetime(' ') == datetime(...), which is
-    # not what we want.  So we first remove all columns that have empty or
-    # whitsepace-only strings to compensate.  This prevents datetime coercion
-    # on whitespace or empty strings.
-    empty_or_ws_cols = df2.apply(lambda s: s.str.strip() == '').all()
-    for colname, remove in empty_or_ws_cols.iteritems():
-        if remove:
-            del df2[colname]
-    df3 = df2.apply(partial(pd.to_datetime, errors='ignore'))
-    for c in df3.columns:
-        df[c] = df3[c]
+    def patched_to_datetime(s):
+        # NOTE: In pandas < 0.17, pd.to_datetime(' ') == datetime(...), which
+        # is not what we want.  This patched version detects empty or
+        # whitespace-only strings, and maps them to a dummy object.  This
+        # prevents datetime coercion on whitespace or empty strings.
+        o = object()
+        s_norm = s.apply(lambda x: o if (hasattr(x, 'strip') and not x.strip()) else x)
+        res = pd.to_datetime(s_norm, errors='ignore')
+        return res.apply(lambda x: '' if x is o else x)
+
+    df2 = df.select_dtypes(include=['object']).apply(patched_to_datetime)
+    for c in df2.columns:
+        df[c] = df2[c]
     return df
 
 
--- a/odo/backends/pandas.py
+++ b/odo/backends/pandas.py
@@ -58,19 +58,18 @@ def coerce_datetimes(df):
     dtype: object
     """
 
-    def patched_to_datetime(s):
-        # NOTE: In pandas < 0.17, pd.to_datetime(' ') == datetime(...), which
-        # is not what we want.  This patched version detects empty or
-        # whitespace-only strings, and maps them to a dummy object.  This
-        # prevents datetime coercion on whitespace or empty strings.
-        o = object()
-        s_norm = s.apply(lambda x: o if (hasattr(x, 'strip') and not x.strip()) else x)
-        res = pd.to_datetime(s_norm, errors='ignore')
-        return res.apply(lambda x: '' if x is o else x)
-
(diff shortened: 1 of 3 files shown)


odo/backends/pandas.py
```diff
@@ -57,9 +57,18 @@ def coerce_datetimes(df):
...
-        partial(pd.to_datetime, errors='ignore')
-    )
+
+    def patched_to_datetime(s):
+        # NOTE: In pandas < 0.17, pd.to_datetime(' ') == datetime(...), which
+        # is not what we want.  This patched version detects empty or
+        # whitespace-only strings, and maps them to a dummy object.  This
+        # prevents datetime coercion on whitespace or empty strings.
+        o = object()
+        s_norm = s.apply(lambda x: o if (hasattr(x, 'strip') and not x.strip()) else x)
```
jreback: use s.str.strip() is much faster that using apply

kwmsmith: Holy crap, that's awesome.  Thanks for the pointer.

jreback: also don't apply pd.to_datetime it works the other way
eg pd.to_datetime(s) is vectorized

odo/backends/pandas.py
```diff
@@ -57,9 +57,20 @@ def coerce_datetimes(df):
     name            object
     dtype: object
     """
-    df2 = df.select_dtypes(include=['object']).apply(
-        partial(pd.to_datetime, errors='ignore')
-    )
+
+    def patched_to_datetime(s):
```
cpcloud: I think a cleaner way to do this is something like the following:

```python
objects = df.select_dtypes(include=['object'])
df2 = objects[[c for c in objects.columns if not np.any(objects[c].str.isspace() | objects[c].str.isalpha())]].apply(partial(pd.to_datetime, errors='ignore'))
```

cpcloud: This is the same logic that we use for inferring datetime values in `datashape.discover`

odo/backends/pandas.py
```diff
@@ -57,9 +57,20 @@ def coerce_datetimes(df):
...
+        # whitespace-only strings, and maps them to a dummy object.  This
+        # prevents datetime coercion on whitespace or empty strings.
+        o = object()
+        try:
+            s[s.str.strip() == ''] = o
+        except TypeError:
+            pass
+        return pd.to_datetime(s, errors='ignore')
+
+    df2 = df.select_dtypes(include=['object']).apply(patched_to_datetime)
```
jreback: you ought to tests this. This i *very* inefficient. ``pd.to_datetime`` is by definition a vectorized operation ITSELF.

kwmsmith: For pandas < 0.17, `to_datetime` is not vectorized on dataframes, only on timeseries.  I checked.

So this is the best we can do, vectorization-wise, for Pandas < 0.17, unless you know something I don't.

kwmsmith:
@jreback @cpcloud I think this is ready to go -- thoughts?

jreback:
@kwmsmith ``pd.to_datetime`` is has always been vectorized by definition. It is ONLY applicable to Series.

you are applying to values!

e.g. doing something like ``df[[.....]].apply(pd.to_datetime)`` is ok as you are applying column by column
by ``df[column].apply(pd.to_datetime)`` is a no-no

cpcloud:
@jreback @kwmsmith isn't apply to values we're doing something similar to your `df[[...]].apply` example

jreback:
oh ok then!

odo/backends/tests/test_csv.py
```diff
@@ -65,6 +65,13 @@ def test_pandas_read_supports_datetimes():
         assert df.dtypes['when'] == 'M8[ns]'
 
 
+def test_pandas_read_supports_whitespace_strings():
```
cpcloud: i don't think this test fails on current master

cpcloud: i just checked out the PR and reverted the fix and this test still passes

kwmsmith: What version of Pandas?  If I revert the fix and run the tests with Pandas `0.16.2`, it fails, as it should.

Pandas 0.17 fixes the `pd.to_datetime()` bug that we're working around here.

```
❯❯❯ py.test odo/backends/tests/test_csv_ws.py
========================================================= test session starts =========================================================
platform darwin -- Python 2.7.10, pytest-2.8.1, py-1.4.30, pluggy-0.3.1
rootdir: /Users/ksmith/work/odo, inifile:
collected 41 items

odo/backends/tests/test_csv.py ....F....................................

============================================================== FAILURES ===============================================================
____________________________________________ test_pandas_read_supports_whitespace_strings _____________________________________________

    def test_pandas_read_supports_whitespace_strings():
        with filetext('a,b, \n1,2, \n2,3, \n', extension='csv') as fn:
            csv = CSV(fn)
            ds = discover(csv)
>           assert ds == datashape.dshape("var * {a: int64, b: int64, '': ?string}")
E           assert dshape("var *...: ?datetime}") == dshape("var * ...'': ?string}")
E             Use -v to get the full diff

odo/backends/tests/test_csv.py:72: AssertionError
================================================= 1 failed, 40 passed in 0.36 seconds =================================================
```

cpcloud:
right ... sorry about that

cpcloud:
merge when ready

Issue #343, opened 2015-10-25
Add Multipart Upload Support for S3


hussainsultan:
upload large files to S3 in smaller chunks.

hussainsultan:
@mrocklin @cpcloud any thoughts here?

mrocklin:
@cpcloud was working on something like this a long while ago.  I think that he's on vacation and getting back in the next day or two.  I plan to leave this to his better judgment.

hussainsultan:
Sounds good. Thank you!

PR #344, opened 2015-10-25
[#343] Add S3 Multipart upload support to aws backend

hussainsultan:
- New S3 Multipart integration tests tries to upload 5MB files due to S3 chunk size limitation. As a result, the test are going to get really slow
- Introduced use_s3_multipart_upload flag as kwargs. Another option is to use file size for perhaps more intelligent default behavior.Please let me know what you guys think.

Also consider,
- Using [Moto](https://github.com/spulec/moto) or other Fake S3 libraries for integration tests that can run locally easily
- Using mocks for unittest
- What could be a good approach to handle default S3 chunk size. Currently, it is hard coded at 5MB.

### DIFF
--- a/odo/backends/aws.py
+++ b/odo/backends/aws.py
@@ -237,12 +237,12 @@ def start_multipart_upload_operation(s3):
 @append.register(S3(JSON), (JSON, Temp(JSON)))
 @append.register(S3(CSV), (CSV, Temp(CSV)))
 @append.register(S3(TextFile), (TextFile, Temp(TextFile)))
-def append_text_to_s3(s3, data, multipart=False, part_size=5 * 1 << 20, **kwargs):
+def append_text_to_s3(s3, data, multipart=False, part_size=5 << 20, **kwargs):
     if multipart:
         with start_multipart_upload_operation(s3) as multipart_upload:
             with open(data.path, 'rb') as f:
-                for part_number, part in enumerate(iter(curry(f.read, part_size), '')):
-                    multipart_upload.upload_part_from_file(BytesIO(part), part_num=part_number + 1)
+                for part_number, part in enumerate(iter(curry(f.read, part_size), ''), start=1):
+                    multipart_upload.upload_part_from_file(BytesIO(part), part_num=part_number)
         return s3
 
     s3.object.set_contents_from_filename(data.path)
--- a/odo/backends/aws.py
+++ b/odo/backends/aws.py
@@ -225,9 +225,10 @@ def start_multipart_upload_operation(s3):
     try:
         multipart_upload = s3.object.bucket.initiate_multipart_upload(s3.key)
         yield multipart_upload
-    except Exception as exception:
-        multipart_upload.cancel_multipart_upload()
-        raise exception
+    except Exception:
+        for part in multipart_upload:
+            s3.object.bucket.cancel_multipart_upload(part.key_name, part.id)
+        raise
     else:
         multipart_upload.complete_upload()
 
@@ -236,26 +237,16 @@ def start_multipart_upload_operation(s3):
 @append.register(S3(JSON), (JSON, Temp(JSON)))
 @append.register(S3(CSV), (CSV, Temp(CSV)))
 @append.register(S3(TextFile), (TextFile, Temp(TextFile)))
-def append_text_to_s3(s3, data, multipart=False, **kwargs):
+def append_text_to_s3(s3, data, multipart=False, part_size=5 * 1 << 20, **kwargs):
     if multipart:
         with start_multipart_upload_operation(s3) as multipart_upload:
-            _upload(data, multipart_upload, **kwargs)
+            with open(data.path, 'rb') as f:
+                for part_number, part in enumerate(iter(curry(f.read, part_size), '')):
+                    multipart_upload.upload_part_from_file(BytesIO(part), part_num=part_number + 1)
         return s3
 
     s3.object.set_contents_from_filename(data.path)
     return s3
-
-
-def _upload(data, multipart_upload, part_size=5 * 1 << 20, **kwargs):
-    parts_counter = 0
-
-    def upload_part(part):
-        multipart_upload.upload_part_from_file(BytesIO(part), part_num=parts_counter + 1)
-        return parts_counter + 1
-
-    with open(data.path, 'rb') as f:
-        for part in iter(curry(f.read, part_size), ''):
-            parts_counter = upload_part(part)
 
 
 try:
--- a/odo/backends/aws.py
+++ b/odo/backends/aws.py
@@ -8,6 +8,7 @@ import re
 from contextlib import contextmanager
 from collections import Iterator
 from operator import attrgetter
+from io import BytesIO
 
 import pandas as pd
 from toolz import memoize, first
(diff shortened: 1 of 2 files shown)


hussainsultan:
@mrocklin @cpcloud any thoughts here?

odo/backends/aws.py
```diff
@@ -219,15 +220,48 @@ def anything_to_s3_text(s3, o, **kwargs):
...
+
 @append.register(S3(JSONLines), (JSONLines, Temp(JSONLines)))
 @append.register(S3(JSON), (JSON, Temp(JSON)))
 @append.register(S3(CSV), (CSV, Temp(CSV)))
 @append.register(S3(TextFile), (TextFile, Temp(TextFile)))
-def append_text_to_s3(s3, data, **kwargs):
+def append_text_to_s3(s3, data, use_s3_multipart_upload=False, **kwargs):
+    if use_s3_multipart_upload:
+        with start_multipart_upload_operation(s3) as multipart_upload:
+            _multipart_upload(data, multipart_upload)
```
cpcloud: if this fails do we still want to call `multipart_upload.complete_upload()`?

odo/backends/aws.py
```diff
@@ -219,15 +220,48 @@ def anything_to_s3_text(s3, o, **kwargs):
...
+    finally:
+        multipart_upload.complete_upload()
+
+
 @append.register(S3(JSONLines), (JSONLines, Temp(JSONLines)))
 @append.register(S3(JSON), (JSON, Temp(JSON)))
 @append.register(S3(CSV), (CSV, Temp(CSV)))
 @append.register(S3(TextFile), (TextFile, Temp(TextFile)))
-def append_text_to_s3(s3, data, **kwargs):
+def append_text_to_s3(s3, data, use_s3_multipart_upload=False, **kwargs):
```
cpcloud: i think a shorter name like `multipart=False` here would be my preference, but i don't feel that strongly about it

odo/backends/aws.py
```diff
@@ -219,15 +220,48 @@ def anything_to_s3_text(s3, o, **kwargs):
...
+
     s3.object.set_contents_from_filename(data.path)
     return s3
 
 
+def _multipart_upload(data, multipart_upload):
+    chunk, chunk_bytes, parts_counter = [], 0, 0
+
+    def upload_part():
+        part = BytesIO(b"".join(chunk))
```
cpcloud: i find the reliance on python's scoping rules here a bit confusing to read .... can you make `chunk` an input parameter and call it like this `upload_part(chunk)` as you build up `chunk` in the loop below?

cpcloud:
@hussainsultan I'm a little skeptical of the benefits of multipart uploading. Won't the network ultimately be the bottleneck here? If I upload a file in 10 parts or 1, ultimately they all have to go through the same network. I could completely be missing a use case here as well ... thoughts?

odo/backends/aws.py
cpcloud: it might be fun to write a function with an API like this:

```python
for chunk in bytestream(data, chunksize=5 * 1 << 20):
    sizeof(chunk) == 5MB
```

cpcloud: actually, you could probably do it with this:

```python
with open(filename, 'rb') as f:
    for chunk in iter(curry(f.read, nbytes), ''):
        yield chunk
```

hussainsultan: ahh! awesome, thanks. didnt think about toolz/curry. I will add this.

hussainsultan:
@cpcloud I agree that this PR itself does'nt add a lot of benefits for uploading. However, i see this as a building block for adding the following features in future:
- Parallel chunk uploads - Useful for high bandwidth network 
- Ability to resume failed uploads - i am not sure how we could add the state of upload in Odo but could be cool

We have been uploading large CSV files with Luigi that has multipart support and it has worked out really well. I'd love to use Odo instead and wanted to make sure that multipart feature is added. Let me know your thoughts.

cpcloud:
I guess I'm not totally sold in the parallel uploads though I'm +1 on resuming failed uploads.

My thinking is this: if I have a high bandwidth network, why wouldn't I just dump the whole file into s3 full stop. What benefits over that does multipart uploading provide?

odo/backends/aws.py
hussainsultan: Good catch, it should be multipart_upload.cancel_multipart_upload()

hussainsultan:
Also, there is some streaming use-cases that could use multipart as well! currently we write a stream to disk on an EC2 instance and then send it over to S3 bucket once the stream is exhausted, it could be cool if we could let a data pass through in a streaming fashion; multipart upload doenst need to know the total file size.

hussainsultan:
@cpcloud I added the requested changes to the pull request. Thank you!

odo/backends/aws.py
```diff
@@ -219,15 +220,44 @@ def anything_to_s3_text(s3, o, **kwargs):
...
 
 
+@contextmanager
+def start_multipart_upload_operation(s3):
+    try:
+        multipart_upload = s3.object.bucket.initiate_multipart_upload(s3.key)
+        yield multipart_upload
+    except Exception as exception:
+        multipart_upload.cancel_multipart_upload()
+        raise exception
```
cpcloud: just reraise here with `raise` instead of `raise exception`:

```python
try:
    ...
except Exception:
    ...
    raise
```

this indicates that you don't intend to handle it as well as removes an unecessary stack frame from the error message

odo/backends/aws.py
```diff
@@ -219,15 +220,44 @@ def anything_to_s3_text(s3, o, **kwargs):
...
 
+def _upload(data, multipart_upload, part_size=5 * 1 << 20, **kwargs):
+    parts_counter = 0
+
+    def upload_part(part):
+        multipart_upload.upload_part_from_file(BytesIO(part), part_num=parts_counter + 1)
+        return parts_counter + 1
+
+    with open(data.path, 'rb') as f:
+        for part in iter(curry(f.read, part_size), ''):
```
cpcloud: i think this can be simplified even further:

```python
for part_num, part in enumerate(iter(curry(f.read, part_size), '')):
    upload_part(part, part_num=part_num)
```
then you wouldn't need this somewhat complex scope logic to increment the parts_counter

cpcloud: even better, you don't need to define a function inside the `_upload` function:

```python
with open(data.path, 'rb') as f:
    for part_num, part in enumerate(map(BytesIO, iter(curry(f.read, part_size), ''))):
        multipart_upload.upload_part_from_file(part, part_num=part_num)
```

hussainsultan:
@cpcloud  thanks for awesome suggestion. I added  it in the latest commit. I am a little weary of the tests right now. I do not have a good way to test the case when upload fails. Do you have any recommendations? I am considering using Moto. Do you have any thoughts?

hussainsultan:
@cpcloud Any concerns before this pull request could be merged?

odo/backends/aws.py
```diff
@@ -219,11 +220,31 @@ def anything_to_s3_text(s3, o, **kwargs):
...
 @append.register(S3(JSONLines), (JSONLines, Temp(JSONLines)))
 @append.register(S3(JSON), (JSON, Temp(JSON)))
 @append.register(S3(CSV), (CSV, Temp(CSV)))
 @append.register(S3(TextFile), (TextFile, Temp(TextFile)))
-def append_text_to_s3(s3, data, **kwargs):
+def append_text_to_s3(s3, data, multipart=False, part_size=5 * 1 << 20, **kwargs):
+    if multipart:
+        with start_multipart_upload_operation(s3) as multipart_upload:
+            with open(data.path, 'rb') as f:
+                for part_number, part in enumerate(iter(curry(f.read, part_size), '')):
```
cpcloud: can you make this `enumerate(..., start=1)`, then you don't need `part_number + 1`?

odo/backends/aws.py
```diff
@@ -219,11 +220,31 @@ def anything_to_s3_text(s3, o, **kwargs):
...
+    else:
+        multipart_upload.complete_upload()
+
+
 @append.register(S3(JSONLines), (JSONLines, Temp(JSONLines)))
 @append.register(S3(JSON), (JSON, Temp(JSON)))
 @append.register(S3(CSV), (CSV, Temp(CSV)))
 @append.register(S3(TextFile), (TextFile, Temp(TextFile)))
-def append_text_to_s3(s3, data, **kwargs):
+def append_text_to_s3(s3, data, multipart=False, part_size=5 * 1 << 20, **kwargs):
```
cpcloud: `part_size=5 * 1 << 20` can be reduced to `5 << 20`

cpcloud:
@hussainsultan some minor comments, otherwise looks good to me

cpcloud:
@hussainsultan i can merge this in today and make that change my self

cpcloud:
oh cool, i see that you've made those changes

PR #345, opened 2015-10-27
Dshape sanitization issue

cpcloud:


odo/backends/tests/test_sql.py
```diff
@@ -716,3 +716,12 @@ def test_numeric_append():
...
     assert discover(sa.REAL()) == float32
+
+
+def test_string_dshape_doc_example():
+    x = np.zeros((10, 2))
+    with tmpfile('.db') as fn:
+        t = odo(
+            x, 'sqlite:///%s::x' % fn, dshape='var * {a: float64, b: float64}'
+        )
+        assert [row == (0, 0) for row in t.select().execute().fetchall()]
```
llllllllll: this could be a listof all `False`, maybe `assert all(...)`

cpcloud: ah duh thanks!

odo/into.py
```diff
@@ -1,23 +1,40 @@
...
+
+_into = namespace['_into']
 
 
-@into.register(type, object)
+@curry
+def into(a, b, dshape=None, **kwargs):
+    if isinstance(dshape, (str, unicode)):
+        dshape = datashape.dshape(dshape)
+    assert isinstance(dshape, datashape.DataShape) or dshape is None, \
```
llllllllll: could this be a typeerror? imho typeerror better conveys that the user passed the incorrect type to the function

cpcloud: yep good idea ... a user could still potentially trigger this

llllllllll:
Other than the two comments, lgtm

cpcloud:
gr8 thx for the review

PR #350, opened 2015-11-06
minor doc add in odo.odo

jdmcbr:
It looks like the description of valid options for `source` had lopped off a line. Just added a reasonable conclusion, and a close parenthesis in the `target` description.


cpcloud:
thx merging now ... those failures are not you .... probably a version change ... investigating now.

PR #353, opened 2015-11-09
Add some support for dask.dataframe

cowlicks:


odo/__init__.py
```diff
@@ -63,6 +63,8 @@
     from .backends.sparksql import SparkDataFrame
 with ignoring(ImportError):
     from .backends.url import URL
+with ignoring(ImportError):
+    from .backends.dask import dask
```
cowlicks: This is the correct idiom for loading a backend right?

mrocklin:
This looks good.  This could be extended by marking `dask.dataframe` as an out-of-core node and then by connecting it to other out of core nodes.

For example in the case of `CSV -> dask.dataframe` we would have to move through `pd.DataFrame`, an in-memory node.

mrocklin:
That's not to suggest that that work should happen now.  This looks like a great start and something that could be merged without the notes from above.

cowlicks:
Anyone know why travis-ci is failing? It segfaults after py.test 'collects' 22 tests. https://travis-ci.org/blaze/odo/jobs/90127217#L556

cowlicks:
Done.

odo/backends/dask.py
```diff
@@ -91,3 +94,21 @@ def bag_to_iterator(x, **kwargs):
...
 def bag_to_iterator(x, **kwargs):
     return db.from_sequence(x, **filter_kwargs(db.from_sequence, kwargs))
+
+
+@convert.register(pd.DataFrame, dd.DataFrame, cost=10)
+def dask_to_pandas_dataframe(x, **kwargs):
+    return x.compute()
+
+
+@convert.register(pd.Series, dd.Series, cost=10)
```
mrocklin: I believe (not sure) that you can stack these

```python
@convert.register(...)
@convert.register(...)
def ...
```

mrocklin:
looks fine to me

cpcloud:
@cowlicks can you add a small release note to `docs/source/whatsnew/0.4.0.txt`? then good to merge

cowlicks:
This should go in for the release. Some unrelated tests are failing in travis-ci.

PR #366, opened 2015-11-23
pandas discovery of float

llllllllll:
`float` types are not optional in pandas. They are properly modelled in the domain of `float32` or `float64`. Until we have option blocks these should not discover as option types.

Also, if a dataframe column reports that it is optional, discovering just that column should not drop the optionality

### DIFF
--- a/docs/source/whatsnew/0.4.0.txt
+++ b/docs/source/whatsnew/0.4.0.txt
@@ -43,6 +43,10 @@ API Changes
 
 * Removed support for Python 2.6 (:issue:`333`).
 * Removed support for Python 3.3 (:issue:`335`).
+* ``pandas.DataFrame`` no longer discovers float32 and float64 as option types
+  (:issue:`366`).
+* ``pandas.Series`` now discovers strings and datetimes as option types
+  (:issue:`366`).
 
 Bug Fixes
 ---------
--- a/odo/backends/pandas.py
+++ b/odo/backends/pandas.py
@@ -4,7 +4,7 @@ from datetime import datetime
 from functools import partial
 
 from datashape import discover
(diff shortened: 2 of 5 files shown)


cpcloud:
Reviewing this now. Going to test against blaze and if passes then I'll merge

cpcloud:
This makes a blaze test fail but I'll fix it after merging this

asaglimbeni:
The Optional is need for missing values in a columns with NaN values. 
This commit broke other behavior for instance the SQL method dshape_to_alchemy now create columns with nullable at false https://github.com/blaze/odo/blob/master/odo/backends/sql.py#L405 even if the serie have missing values that in SQL means NULL values. So there are two options or rollback or you have to fix even in the other drivers.

llllllllll:
NaN is not missing for float, it is a valid element of float. Optional float is the set containing all floats (including NaN) and one new NULL value. You can store NaN in a non nullable float column:

```
test=# select * from floattest ;
  a  
-----
    
    
    
   1
   2
   3
 NaN
 NaN
 NaN
 0.0
 0.0
 0.0
(12 rows)
```

this column has valid floats (including `NaN`) and also `NULL`

asaglimbeni:
Not all the DBs accept NaN as valid value in non nullable column. 
And in any case the append_iterator_to_table https://github.com/blaze/odo/blob/master/odo/backends/sql.py#L449 don't do that because leave free sqlAlchemy to build the insert stantment to the DB. 
The case is the follow if you try to use odo to write a dataframe (with nan values in float columns) to the DB like MSSQL it creates float columns with nullable at false and when it try to insert row with NaN the DBe engine rises becuase it is not allowed. 
I guess we need to refactor a litle the append_iterator_to_table to be coherent with this behaivor.

PR #371, opened 2015-11-30
Fix quoting of MySQL table name

cpcloud:
Fixes #316

### DIFF
--- a/odo/backends/sql_csv.py
+++ b/odo/backends/sql_csv.py
@@ -84,14 +84,21 @@ def compile_from_csv_sqlite(element, compiler, **kwargs):
     cmd = ['sqlite3',
            '-nullvalue', repr(element.na_value),
            '-separator', element.delimiter,
-           '-cmd', '.import "%s" %s' % (fullpath, t.name),
+           '-cmd', '.import "%s" %s' % (
+               # FIXME: format_table(t) is correct, but sqlite will complain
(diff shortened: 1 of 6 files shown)


cpcloud:
@llllllllll any comments here? else merging on pass

odo/backends/sql_csv.py
```diff
@@ -84,65 +84,97 @@ def compile_from_csv_sqlite(element, compiler, **kwargs):
...
-    assert not stdout, \
-        'error: %s from command: %s' % (stdout, ' '.join(cmd))
+    stdout, stderr = subprocess.Popen(
+        cmd,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+        stdin=subprocess.PIPE
+    ).communicate()
+    if stdout:
+        raise RuntimeError(
```
llllllllll: Could we raise something more specific here, I normally think of runtime error as recursion depth and notimplemented style issues.

llllllllll:
Other than the one comment this looks good. I can close the other pr I have open about encodings also.

odo/backends/sql_csv.py
cpcloud: Yeah I'll raise a sqlalchemy exception here instead.

llllllllll: thanks

PR #377, opened 2015-12-08
Remove dynd from conda.recipe

cowlicks:
dynd is currently unstable so do not test on it.


cowlicks:
This was breaking the conda.recipe merging soon if no comment

PR #381, opened 2015-12-15
pip install dask[complete] in travis-ci

cowlicks:


cowlicks:
Yay it worked. Merging.

PR #383, opened 2015-12-17
Tweak dask backend costs. Add a default chunks size for dask.

cowlicks:
When odo tries to route through the current dask.array backend it does not set the `chunks` keyword arg, so it fails.

This came up while preparing the blaze release today.

### DIFF
--- a/odo/backends/dask.py
+++ b/odo/backends/dask.py
@@ -53,10 +53,18 @@ else:
         return x.resize(size)
 
 
-@convert.register(Array, tuple(arrays), cost=1.)
+@convert.register(Array, tuple(arrays), cost=3.0)
 def array_to_dask(x, name=None, chunks=None, **kwargs):
     if chunks is None:
-        raise ValueError("chunks cannot be None")
+        # chunk into chunks of 1,000 elements along each axis.
+        cs = 1e3
+        chunks = []
+        for i in x.shape:
+            if i > cs:
+                main = (int(cs),) * int((i//cs))
+                rem = (int(i % cs),) if (i % cs) else ()
+                chunks.append(main + rem)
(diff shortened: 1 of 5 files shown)


cpcloud:
This was never a problem before. This is a regression and should be fixed by figuring out why we are now going through dask when before we weren't

cowlicks:
@cpcloud I think it *is* a bug that when odo routes through `array_to_dask`, it does not set the `chunks` keyword argument, which raised a `ValueError` which made the conversion fail.

But, it is a regression that odo is now routing through this path when it wasn't before. I'm not sure why this is happening. And running the blaze tests today the failure does not happen anymore >____<

odo/backends/dask.py
```diff
@@ -53,10 +53,18 @@ def resize(x, size):
         return x.resize(size)
 
 
-@convert.register(Array, tuple(arrays), cost=1.)
+@convert.register(Array, tuple(arrays), cost=3.0)
 def array_to_dask(x, name=None, chunks=None, **kwargs):
     if chunks is None:
-        raise ValueError("chunks cannot be None")
+        # chunk into chunks of 1,000 elements along each axis.
+        cs = 1e3
```
cpcloud: this seems a little arbitrary, what's wrong with enforcing that the user provides chunks? not providing a default chunks if we're going through dask unintentionally is also nice because then we get the error message

cowlicks: It is definitely an arbitrary heuristic. Should a user have to provide a `chunks` keyword argument even if they don't know what it is?

cpcloud: Dask itself requires chunks, so shouldn't we follow its lead?

cpcloud:
Should this PR be closed in favor of #384?

PR #384, opened 2015-12-17
Fix cost of converting a Select to a scalar value

cpcloud:


cpcloud:
@cowlicks this fixes the issue properly which was that the cost of getting a scalar out of a select statement from sqlalchemy was actually higher than getting out a series or dataframe.

cowlicks:
@cpcloud Do you think, the costs for `dask_to_other` should be higher? I thought they should be comparable with `select_to_base`?

cpcloud:
@cowlicks Yep, that makes sense. I'll update accordingly.

cowlicks:
Thanks @cpcloud

PR #328, opened 2015-09-28
Allow connect_args by removing memoization of sa.create_engine

cpcloud:

PR #385, opened 2015-12-18
don't use memoized metadata for SqlAlchemy, resolves #380

### LINKED ISSUE #380: Using drop to overwrite data to postgresql
I am having a very inconsistent result when trying to do `odo()` to postgresql target. As I can't find a way to overwrite existing record, I use `drop()` before calling `odo()`.

```
drop('postgresql://.....')
odo(dataframe, 'postgresql://....')
```

Then I start getting exceptions because the table does not exist, hence I did this

```
try:
    drop('postgresql://....')
except ValueError:
    pass

odo(dataframe, 'postgresql://....')
```

So when I ran the code above the first time, I got exception for `odo()` saying my table does not exist. However, when I ran the same code the second time, the exception didn't get raised again, and the newly created table was properly populated. Am I doing anything wrong?

I am on odo 0.3.4 and python 3.4.3 on ubuntu 14.04.3

branning:
issue #380 occurs because the SqlAlchemy metadata for the engine is only retrieved once, due to the memoization of `metadata_of_engine`. Since the engine object isn't changed, the cached metadata is re-used, even though a table has been dropped and the metadata should be different.

In my opinion, it's not safe to memoize the SqlAlchemy metadata in any case. It's not a pure function. It seems like this kind of issue might happen often.

### DIFF
--- a/odo/backends/sql.py
+++ b/odo/backends/sql.py
@@ -515,7 +515,7 @@ def resource_sql(uri, *args, **kwargs):
     # we were also given a table name
     if args and isinstance(args[0], (str, unicode)):
         table_name, args = args[0], args[1:]
-        metadata = metadata_of_engine(engine, schema=schema)
+        metadata = sa.MetaData(engine, schema=schema)
 
         with ignoring(sa.exc.NoSuchTableError):
             return attach_schema(


llllllllll:
Will this cause us to create a lot of extra connections each time we call `resource`?

branning:
tThe failed test is [the test of metadata caching](https://github.com/blaze/odo/blob/master/odo/backends/tests/test_sql.py#L372) [(source)](https://github.com/blaze/odo/blob/master/odo/backends/tests/test_sql.py#L372), which should fail without the caching.

@llllllllll the same engine is used for each call to `resource`, which is why the memoization was a problem. But I don't know if this will affect connection reuse.

cpcloud:
@branning I've fixed this for a slightly different issue here: https://github.com/blaze/odo/pull/328 (I wanted to pass in an argument to create engine that isn't easily memoized). Can you give this a go and see if it works for you? Will merge on +1 from you and passing on travis

branning:
@cpcloud the change I made is a subset of your pull request, so I expect it to also resolve #380. I'm getting that code now and I'll verify that it also works. I expect the same metadata caching test will fail on Travis.

cpcloud:
@branning The test shouldn't fail there because I've corrected it to reflect the changes :)

branning:
Ah, good job! I see that now, insert a `not` and the test now works the other way. Pull request #328 also fixed the problem in #380 for me.

PR #408, opened 2016-01-21
Use docker to provision mongo and postgres servers for testing

kwmsmith:
@sandhujasmine: you should be able to do the following (assuming a working `docker`):

```
# do this once -- launches and runs two docker containers, and sets up environment variables.
$ eval $(provisioning.sh)  
$ py.test
$ conda build conda.recipe
```

let me know if you have issues.


sandhujasmine:
I believe I have the docker containers running the DBs. I started the docker-machine (via quickstart terminal). Then ran the provisioning script. As a check, if I run it twice, it tells me the port is in use so assume docker containers are running mongo and postgres correctly. 

Looks like all tests are passing. However, if I stop the docker-machine, all odo tests still pass so assume it must not hit the DB. Investigating further and will spend the afternoon learning through this and looking into source code ..

```
py.test odo/tests/
======================================================== test session starts =========================================================
platform darwin -- Python 3.5.1, pytest-2.8.5, py-1.4.31, pluggy-0.3.1
rootdir: /Users/jsandhu/Documents/projects/odo, inifile: 
collected 83 items 

odo/tests/test_append.py ....
odo/tests/test_chunks.py ......
odo/tests/test_convert.py ................................
odo/tests/test_core.py ...
odo/tests/test_directory.py ..
odo/tests/test_into.py ...........
odo/tests/test_numpy_dtype.py ..............
odo/tests/test_pydot.py .
odo/tests/test_regex.py .
odo/tests/test_resource.py .
odo/tests/test_temp.py .
odo/tests/test_utils.py ......
odo/tests/test_version.py .
```

odo/backends/tests/test_postgres.py
```diff
@@ -23,35 +23,46 @@
...
 
 
 @pytest.yield_fixture
-def complex_csv():
+def complex_csv(tmpdir):
     path = os.path.join(os.path.dirname(__file__), 'dummydata.csv')
-    with tmpfile('.csv') as fn:
+    with tmpfile('.csv', dir=tmpdir) as fn:
         shutil.copy(path, fn)
+        os.chmod(fn, 0777)
```
llllllllll: This is a syntax error in py3, use `0o777` which is py2/3 comptible and imo more clear.

kwmsmith: Good catch, thanks.

sandhujasmine:
In testing this, several tests in test_postgres.py are failing. Looking at one of them: `test_sql_to_csv()`, it looks like the failure lies in [compile_copy_to_csv_postgres](https://github.com/blaze/odo/blob/master/odo/backends/sql.py#L632)

The original error was Permission Denied:
```

self = <sqlalchemy.dialects.postgresql.psycopg2.PGDialect_psycopg2 object at 0x105acc3d0>, cursor = <cursor object at 0x10586d430; closed: -1>
statement = 'COPY tbl0 TO %(path)s\n            WITH (\n                FORMAT CSV,\n                HEADER %(header)s,\n         ...a_value)s,\n                ESCAPE %(escapechar)s,\n                ENCODING %(encoding)s\n            )\n            '
parameters = {'delimiter': ',', 'encoding': 'UTF8', 'escapechar': '\\', 'header': True, ...}
context = <sqlalchemy.dialects.postgresql.psycopg2.PGExecutionContext_psycopg2 object at 0x105d60810>

    def do_execute(self, cursor, statement, parameters, context=None):
>       cursor.execute(statement, parameters)
E       ProgrammingError: (psycopg2.ProgrammingError) could not open file "/Users/jsandhu/tmp/tmpvDnqET.csv" for writing: Permission denied
E        [SQL: 'COPY tbl0 TO %(path)s\n            WITH (\n                FORMAT CSV,\n                HEADER %(header)s,\n                DELIMITER %(delimiter)s,\n                QUOTE %(quotechar)s,\n                NULL %(na_value)s,\n                ESCAPE %(escapechar)s,\n                ENCODING %(encoding)s\n            )\n            '] [parameters: {'escapechar': '\\', 'encoding': u'UTF8', 'header': True, 'delimiter': ',', 'quotechar': '"', 'path': '/Users/jsandhu/tmp/tmpvDnqET.csv', 'na_value': ''}]
```

But in searching for this online, it appears the format of the [SQL statement](http://stackoverflow.com/questions/22527017/psycopg2-is-denied-permissions-on-tmp) is incorrect. So tried updating this but that also didn't work because looks like we have to drop down to underlying psycopg2 driver and use it's `copy_expert() or copy_to()` commands. See following:

http://stackoverflow.com/questions/13125236/sqlalchemy-psycopg2-and-postgresql-copy

https://groups.google.com/forum/?fromgroups=#!topic/sqlalchemy/vZIztYaqlpo

Saw Philip's @cpcloud note here: https://bitbucket.org/zzzeek/sqlalchemy/issues/3404/copy-to-csv-expression

Following works but is a hack and doesn't integrate well with the system since we want to use sqlalchemy's interface with `execute()` .. still trying to figure it out. Would be curious to know if this test passes for others.

```
def test_sql_to_csv_x(sql, csv, tmpdir):
    sql, bind = sql
    sql = odo(csv, sql, bind=bind)
    with tmpfile('.csv', dir=tmpdir) as fn:
        # following fails
        csv = odo(sql, fn, bind=bind)
        # try something else
        conn = bind.raw_connection()
        stmt = "COPY tbl5 TO STDOUT WITH CSV HEADER"
        cur = conn.cursor()
        with open(fn, 'w') as h:
            cur.copy_expert(stmt, h)

        assert odo(csv, list) == data
```

Below are my versions:
```
In [1]: import odo

In [2]: odo.__version__
Out[2]: '0.4.1rc1+9.g181a10c.dirty'

In [3]: import sqlalchemy

In [4]: sqlalchemy.__version__
Out[4]: '1.0.11'

In [5]: import psycopg2

In [6]: psycopg2.__version__
Out[6]: '2.6.1 (dt dec pq3 ext)'
```

kwmsmith:
> Looks like all tests are passing. However, if I stop the docker-machine, all odo tests still pass so assume it must not hit the DB. Investigating further and will spend the afternoon learning through this and looking into source code ..

Great -- one thing to check is whether the tests are skipped when the DB is down.

> In testing this, several tests in test_postgres.py are failing. Looking at one of them: test_sql_to_csv(), it looks like the failure lies in compile_copy_to_csv_postgres

Hmm, OK, all my tests are passing here, I'll look into this.  Likely best if we sync up to see what the difference is.

kwmsmith:
Quoting from email from @sandhujasmine:

>  So, we can decouple the management of the docker script from the testing scripts a little bit.

Great.

> 1. We can use provisioning.sh to manage the docker part. It doesn't need to define POSTGRES_IP or MONGO_IP since python can do a subprocess call to get this info directly. The benefit being we don't have to set the environment outside of python. 
The only catch is the tmp directory must be the same for docker container and pytest. 


This is good, although one thing to keep in mind is that I'd prefer to keep the docker part optional; if a developer wants to manually set up mongo and postgres and run the tests without docker, that should be possible.

> 2. This might be a digression but I ended up playing w/ provisioning script to make it a bit more robust. When we're testing, like today, maybe testing will be expedited and easier for new users if we can do: `provisioning.sh restart mongo|postgres` - just add a bit more functionality to it.

Yes, restarting is good.  Another improvement would be to make it idempotent -- if you run `provisioning.sh start` (or whatever the command is to start it up) twice in a row, the second time is a no-op.

> Maybe I can add a command like option for pytest (in conftest.py something like --docker) so it'll set the environment variables by making the subprocess call; so we don't have to run provisioning.sh before running tests. I thought this would be useful during development and for new users when we might need to restart our docker containers or have multiple terminals open and we don't have to worry about running or sourcing provisions to define environment vars before we can run tests.

That looks promising -- please explore but don't spend too much time on it.

sandhujasmine:
Yes - understand your point that pytest should not need to know where the DB is running. It just needs the IP addresses and the (tmp) dir for write location.

Will try linux this morning. I realize adding option to pytest is also not helpful since the VMs will not be able to do a `docker machine ip` as docker is running on host machine. If we can save config like IP and tmp dir in a file and have both docker + pytest read it from there, that'd be ideal.

But I'll work on testing linux against current docker setup next.

kwmsmith:
OK, this is working as advertised, thanks @sandhujasmine.  Merging.

PR #409, opened 2016-01-22
BUG: start memoizing metadata again

llllllllll:
We are leaking a lot of connections right now which is causing me to hit connection limits. 

This change addresses the bugs that we fixed in #328 by using a frozenset to pass the dict to `create_engine`. We also fix the issue with `drop` by explicitly removing the table from the metadata.

@cpcloud


llllllllll:
@branning Added a commit that does what I actually was trying to do

Issue #412, opened 2016-01-26
Add doc string for resource.register()


dr-leo:
Hi,

I haven't seen any docs on this. I would like to add odo support to pandasdmx. It should be about 3 lines. But without documenting the register API even the simplest task turns into an uphill battle. In particular, what 
are the kwargs good for? Can I pass args to the registered callable? By accident I found out that the kwargs contain a dshape item...

llllllllll:
does the docstring in #413 help?

dr-leo:
I must admit that a quick look does not entirely answer all my
questions. But I cannot take more time to do it thoroughly.

In the main, I'd like to understand what's passed to myfunc:

@odo.resource.register('.*\.mysuffix')
def myfunc(uri, **kwargs):
    # do the conversion. To what format? csv? PD.DataFrame? jpeg? mp3?
Why is DataFrame better than # mp3? How do the kwargs passed help? Cann
the caller controll the conversion such as in ``
odo('somefile.mysuffix', 'targetfile.csv', careful=True, lazy=False)``
    return SomeType # which ones are good?

It would be good to read that kwargs contains dshape (and what else?),
and that the conversion function should ideally return the data in a
format understood by odo such as PD.DataFrame or whatever. The example
on rtd with pickle is not so inspiring because it does not talk about
what a good return type might be, and what the **kwargs are good for.

I do understand that the regex is meant to capture the source file
extension although this is nowhere documented either. It is likewise
unclear whether the passed regex must end with '$' to make sure that
'*.sdmx' and '*.sdmxml' are not treated the same. But it looks like
normalisation does just that...

Sorry for bothering. I'm an odo newbie who does not want to become an
odo pro. I am just a pandasdmx pro who heard about odo, thinks it is
great and wants to spend a few minutes writing 3 lines to register a
conversion...



Am 26.01.2016 um 21:25 schrieb Joe Jevnik:

llllllllll:
> do the conversion. To what format? csv? PD.DataFrame? jpeg? mp3?

This is not actually correct. `resource` does not do any conversions, that is `convert`'s job.
`resource` is only used to return the objects that should appear in the odo graph. This allows us to turn `str` into `sa.Table` so that we can turn a table into a dataframe, or give the path to a csv on disc or s3. The kwargs are what ever you happen to pass to `resource`, so there is no spec for it. The user is the one who calls it. The positional args are the uri, and then maybe a field specified by `::<name>` like `postgresql://localhost/db::table`.

[...]

PR #413, opened 2016-01-26
Regex dispatcher docs

llllllllll:
Fixes the categorical check on older versions of pandas.

Adds docstrings as requested in: #412 
Compiles the patterns once.


kwmsmith:
LGTM.

@dr-leo does this `register` docstring improve things?

PR #414, opened 2016-01-29
ENH: Docker provioning more robust/user friendly # 408

sandhujasmine:
Added start|stop|restart capability per backend (mongo|postgres)
Provide tmp_dir as input as well as tmp_dir mapping for Linux VM

Tested it on Linux + Mac and it works.
Would be good to add a check to ensure TMP_DIR for postgres is read/write by everyone; however, couldn't yet get this working in bash script.

Also somehow managed to corrupt the docker images through repeated testing. Needs more testing w/ various options [start|restart] perhaps? to see if there is an issue here.

`./provisioning.sh -h` provides usage information.

@kwmsmith


kwmsmith:
@sandhujasmine I've merged this branch into #408, and will take it all the way home on that PR.  Closing.

PR #418, opened 2016-02-02
Remove spark 1.2

### LINKED ISSUE #296: deprecate support for spark 1.2
My feeling is that many people moved to 1.3 right after it came out because of the `DataFrame` interface and there are some crufty things we do to support all the spark versions. I'd like to remove them.

kwmsmith:
Closes #296.


llllllllll:
:sparkles:

PR #420, opened 2016-02-04
Fix extra-requirements.txt is missing from source distribution on PyPI

cgohlke:


llllllllll:
Thanks for this!

kwmsmith:
@cgohlke @llllllllll 0.4.2 was tagged and uploaded to PyPI, and I verified it's `pip` installable; apologies for the hassle.

PR #428, opened 2016-02-19
ENH: Add option to mount additional drives

sandhujasmine:
Some of the blaze tests populate postgres data from CSV file in the
blaze/ project's directory. Use this option to mount it in docker so
posgres can read data when running those tests.

### DIFF
--- a/provisioning.sh
+++ b/provisioning.sh
@@ -4,7 +4,7 @@
 
 help(){
   echo "Usage:"
-  echo "   \$ provisioning.sh [-h] -t read_write_loc_on_mac [-lt mnt_on_linux_vm] start|stop|restart [mongo|postgres]"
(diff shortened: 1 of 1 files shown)


sandhujasmine:
This allows us to define another mount point using `-m` so psql can find and load data for blaze tests.

To run blaze tests in test_postgresql_compute.py, let's fire up the postgres docker container and mount the test data location as well as the temp location to write output of tests:

`provisioning.sh -t ~/tmp/docker/ -m ${BLAZE_PROJ_PATH}/blaze/blaze/examples/data restart`

Next, set the env variables in the terminal. provisioning script writes a docker_info.txt file in `~/tmp/docker` - copy & paste first 3 lines in terminal:
```
export POSTGRES_IP=192.168.99.100
export MONGO_IP=192.168.99.100
export POSTGRES_TMP_DIR=/Users/jsandhu/tmp/docker/
```

The following will now run tests that were previously skipping:
`py.test -v test_postgresql_compute.py`

kwmsmith:
+1  Sorry that this one slipped through the cracks.

PR #430, opened 2016-03-02
BUG: This get's MSSQL server working...

thequackdaddy:
Hello,

I wanted to show you what I did to get Blaze working with MSSQL. 

A few points...
* I'm not sure why, but you have a call to `typ.collation` and have that set the encoding of the field. I'm confused by the logic behind this fiend. 
* MSSQL has a few unique datatypes. Chief among them (from my perspective), is BIT. I just hard-added it to the SQL. I admit this is not ideal, and am open to discussing other ways to make it work. I don't know what the correct practice is. 

Once I did this, I could run... 
```
import blaze as bz
import odo
a = bz.Data(odo.resource('mssql+pyodbc://MyDSN::TableName',schema='dbo'))
a.Field.sum()
```

Small complaint... I get a an sqlaclchemy warning... 

```
C:\Anaconda3\lib\site-packages\sqlalchemy\dialects\mssql\base.py:1170: SAWarning: legacy_schema_aliasing flag is defaulted to True; some schema-qualified queries may not function correctly. Consider setting this flag to False for modern SQL Server versions; this flag will default to False in version 1.1
  "legacy_schema_aliasing flag is def
```

Anyone have an idea how to silence this? I tried a few things and couldn't figure it out. 

Thanks for reviewing. I'm happy to amend this to make it fit with your programming practices, but just wanted to show you what worked.

### DIFF
--- a/odo/backends/sql.py
+++ b/odo/backends/sql.py
@@ -18,6 +18,7 @@ from sqlalchemy import inspect
 from sqlalchemy.ext.compiler import compiles
 from sqlalchemy import event
 from sqlalchemy.schema import CreateSchema
+from sqlalchemy.dialects import mssql
 
 from multipledispatch import MDNotImplementedError
 
@@ -84,6 +85,11 @@ revtypes.update({
     sa.Float: float64,
     sa.Float(precision=24): float32,
     sa.Float(precision=53): float64,
+    mssql.BIT: datashape.bool_,
+    mssql.DATETIMEOFFSET: string,
+    mssql.MONEY: float64,
+    mssql.SMALLMONEY: float32,
+    mssql.UNIQUEIDENTIFIER: string
 })
 
 # interval types are special cased in discover_typeengine so remove them from
@@ -172,7 +178,7 @@ def discover_typeengine(typ):
     if isinstance(typ, (sa.NUMERIC, sa.DECIMAL)):
         return datashape.Decimal(precision=typ.precision, scale=typ.scale)
     if isinstance(typ, (sa.String, sa.Unicode)):
-        return datashape.String(typ.length, typ.collation)
+        return datashape.String(typ.length, 'U8')
     else:
         for k, v in revtypes.items():
             if isinstance(k, type) and (isinstance(typ, k) or
--- a/odo/backends/tests/test_sql.py
+++ b/odo/backends/tests/test_sql.py
@@ -281,19 +281,19 @@ def test_discover_oracle_intervals(freq):
 def test_mssql_types():
     typ = sa.dialects.mssql.BIT()
     t = sa.Table('t', sa.MetaData(), sa.Column('bit', typ))
-    assert discover(t) == dshape('var * {bit: ?bool}')
+    assert_dshape_equal(discover(t), dshape('var * {bit: ?bool}'))
     typ = sa.dialects.mssql.DATETIMEOFFSET()
     t = sa.Table('t', sa.MetaData(), sa.Column('dt', typ))
-    assert discover(t) == dshape('var * {dt: ?string}')
+    assert_dshape_equal(discover(t), dshape('var * {dt: ?string}'))
     typ = sa.dialects.mssql.MONEY()
     t = sa.Table('t', sa.MetaData(), sa.Column('money', typ))
-    assert discover(t) == dshape('var * {money: ?float64}')
+    assert_dshape_equal(discover(t), dshape('var * {money: ?float64}'))
     typ = sa.dialects.mssql.SMALLMONEY()
     t = sa.Table('t', sa.MetaData(), sa.Column('money', typ))
-    assert discover(t) == dshape('var * {money: ?float32}')
+    assert_dshape_equal(discover(t), dshape('var * {money: ?float32}'))
     typ = sa.dialects.mssql.UNIQUEIDENTIFIER()
     t = sa.Table('t', sa.MetaData(), sa.Column('uuid', typ))
-    assert discover(t) == dshape('var * {uuid: ?string}')
+    assert_dshape_equal(discover(t), dshape('var * {uuid: ?string}'))
 
 
 def test_create_from_datashape():
--- a/odo/backends/tests/test_sql.py
+++ b/odo/backends/tests/test_sql.py
@@ -278,6 +278,24 @@ def test_discover_oracle_intervals(freq):
     assert discover(t) == dshape('var * {dur: ?timedelta[unit="%s"]}' % freq)
 
 
+def test_mssql_types():
+    typ = sa.dialects.mssql.BIT()
+    t = sa.Table('t', sa.MetaData(), sa.Column('bit', typ))
+    assert discover(t) == dshape('var * {bit: ?bool}')
+    typ = sa.dialects.mssql.DATETIMEOFFSET()
+    t = sa.Table('t', sa.MetaData(), sa.Column('dt', typ))
+    assert discover(t) == dshape('var * {dt: ?string}')
+    typ = sa.dialects.mssql.MONEY()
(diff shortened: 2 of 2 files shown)


branning:
This looks like it might fix errors [like this one reported on Stack Overflow](http://stackoverflow.com/questions/33644815/unsupported-string-encoding-using-blaze-dataengine}

thequackdaddy:
Probably...

I see 2 problems that need fixin':

* In MSSQL (at least) collation != dtype
* MSSQL has a smorgasbord of custom name values. I'm happy to try my hand at adding them to the SQL Alchemy and datashape libraries, but I'm hoping someone can point to the right place to do this. I'm *decent* at python (read that as a step up from "I stayed at a Holiday Inn last night").

odo/backends/sql.py
```diff
@@ -172,7 +172,9 @@ def discover_typeengine(typ):
     if isinstance(typ, (sa.NUMERIC, sa.DECIMAL)):
         return datashape.Decimal(precision=typ.precision, scale=typ.scale)
     if isinstance(typ, (sa.String, sa.Unicode)):
-        return datashape.String(typ.length, typ.collation)
+        return datashape.String(typ.length, None)
```
llllllllll: do youy know if there is a way to get the encoding off the type? I remember looking at this in the past but I don't remember why I didn't fix this. I agree that passing the collation for the encoding is strictly incorrect though.

odo/backends/sql.py
```diff
@@ -172,7 +172,9 @@ def discover_typeengine(typ):
...
+    if isinstance(typ, sa.dialects.mssql.BIT):
+        return datashape.int_
```
llllllllll: I don't know mssql at all but the docs imply that this should be `datashape.bool_`

llllllllll:
Regarding the warning: I would prefer understanding what is actually causing this warning, it sounds like we might need to thread some new arguments around. Can you try to get the traceback from this warning,. A simple way to get this trace would be to turn that warning into an error with:

```python
import warnings

with warnings.catch_warnings():
    warnings.simplefilter('error')
    code_that_triggers_warning()
```

I don't think anyone on the odo team has run with MSSQL so I fear that you may run into more issues like this in the future. Thank you for helping out by providing a fix!

odo/backends/sql.py
thequackdaddy: I feel like there was a reason that I picked that on bool... I can't seem to recall what that was. 

My real point is that there's a lot of unique MSSQL types. See [here](http://docs.sqlalchemy.org/en/latest/dialects/mssql.html#sql-server-data-types). What I'm wondering if I should make a dict with a bunch of mappings to them like type and revtype (lines 48-87 in backends/sql.py) or is this something that we shoudl send to SQL Alchemy? I'm not really sure where/how this transformation *ought* to take place. I'm happy to write the dict, I just want to do it the right place.

odo/backends/sql.py
thequackdaddy: Vocabulary question here... by "encoding" do you mean (this)[https://docs.python.org/3/library/codecs.html#standard-encodings]? I can check.

thequackdaddy:
> I don't think anyone on the odo team has run with MSSQL so I fear that you may run into more issues like this in the future. Thank you for helping out by providing a fix!

Fear not.

odo/backends/sql.py
llllllllll: for things like bit here which do not need to do any conditional construction I think it would be okay to just add this to the `revtypes.update(...)` mapping. To simplify the code path here if we have a lot of types that need some specialization we could use mulipledispatch to do something like:

```python
@dispatch(type)
def from_sql_type(type_):
    if type_ in revtypes:
        return dshape(revtypes[type_])[0]
    # body of the else in the current function
    # for k, v in revtypes.items():
    #     ...

@dispatch(object)
def from_sql_type(ob):
    return _from_sql_type(type(ob))

@dispatch((sa.String, sa.Unicode))
def from_sql_type(type_):
    return datashape.String(type_.length, None)

...
```

This might help us add more special handling in the future without requiring us to make this function a nasty if/else mess. Depending on what we decide we could even document this function to allow users to register their own custom sqlalchemy type converters!

Does this approach make sense or do you think it is overkill for something like this?

odo/backends/sql.py
llllllllll: Yes, something that will allow us to convert `bytes` into `unicode`

odo/backends/sql.py
thequackdaddy: Ok, I'll try to whip something like that up in the next week or so. Thanks.

odo/backends/sql.py
thequackdaddy: Ok after doing a little digging... 

the type type is NVARCHAR or NCHAR, that's by definition unicode. (I think SQL Server uses UTF-16)

If the type is CHAR or VARCHAR, then the code page (which I think is basically the same thing as encoding) is dependent on the collation. 

https://msdn.microsoft.com/en-us/library/ms190305.aspx

http://www.sqlservercentral.com/blogs/rocks/2012/01/09/revised-difference-between-collation-sql_latin1_general_cp1_ci_as-and-latin1_general_ci_as/

So right now, I'm thinking there's gonna have to be some logic that says Collation -> Code Page -> python encoding. Doesn't look fun.

thequackdaddy:
Hello everyone,

Finally got back to this. 

I struggle with getting an encoding from the `typ` value from `sqlalchemy`. Per [this documentation](http://docs.sqlalchemy.org/en/latest/core/engines.html#sqlalchemy.create_engine.params.encoding), it sounds to me like if we are working with python 3, then we should only expect U-8 unicode. (I'm still a little fuzzy on exactly how/what is transmitted from the a generic SQL server to Python. From my experiments, I was always under the impression this was nearly universally already ascii or unicode. So I'm not fully sure how to check for the right encoding. 

I've added a few unit tests, which I don't necessarily think are all that amazing. They test the types, but considering the types are coded, they are kind of ho-hum. If someone has an idea of how to write a unit test for the encoding piece, that would greatly help my ability to check if the stream is ascii or U-8. For the moment, I would suggest assuming everything is U-8. 

Thanks for your review.

thequackdaddy:
It says the build failed, but I think that's a travis-ci problem. Built just fine for me here:

https://travis-ci.org/thequackdaddy/odo/builds/132938026

llllllllll:
Yeah, it timed out doing some IO. it was only on one of the runs in the matrix so I restarted it

odo/backends/tests/test_sql.py
```diff
@@ -278,6 +278,24 @@ def test_discover_oracle_intervals(freq):
     assert discover(t) == dshape('var * {dur: ?timedelta[unit="%s"]}' % freq)
 
 
+def test_mssql_types():
+    typ = sa.dialects.mssql.BIT()
+    t = sa.Table('t', sa.MetaData(), sa.Column('bit', typ))
+    assert discover(t) == dshape('var * {bit: ?bool}')
```
llllllllll: could you change these to `datashape.util.assert_dshape_equal`. This function provides the path to any attributes that are invalid, for example:

```
In [4]: assert_dshape_equal(
    dshape('var * {a: var * var * {b: int32, c: int64}}'),
    dshape('var * {a: var * var * {b: int32, c: int32}}'),
)
---------------------------------------------------------------------------
AssertionError
  ...
AssertionError: ctype names do not match: 'int32' != 'int64'
path: _.measure['a'].measure['c'].name
```

thequackdaddy: Done.

llllllllll:
Merging, thanks!

thequackdaddy:
Thank you.

PR #433, opened 2016-03-17
BUG: SQLAlchemy Column names -- force to be string

### LINKED ISSUE #432: Reading from SQL to pandas result in column names of sqlalchemy type
Hello, 

Take this example:

```
import odo
import pandas as pd
odo.odo('mssql+pyodbc://MyMSSQLServer::MyTable',
        'hdfstore://c:/workspace/MyHDFFile.h5::/MyTable', schema='dbo')

h = pd.HDFStore('c:/workspace/MyHDFFile.h5')
type(h.select('/data', 'index <= 1').MyField.name)
```

This returns a type of `sqlalchemy.sql.elements.quoted_name`. This messes up a program I'm running down the road. I'd like to make it so that this is just forced to be a `str` type. I'm happy to try to code this, but if someone could point at the right files/lines of code to try, I'd appreciate it. I'm not familiar with how this program works.

Thanks.

thequackdaddy:
Hello,

This resolves #432.

Thanks.

### DIFF
--- a/odo/backends/sql.py
+++ b/odo/backends/sql.py
@@ -199,13 +199,31 @@ def discover_sqlalchemy_column(c):
 
 @discover.register(sa.sql.FromClause)
 def discover_sqlalchemy_selectable(t):
-    ordering = dict((c, i) for i, c in enumerate(c for c in t.columns.keys()))
-    records = list(sum([discover(c).parameters[0] for c in t.columns], ()))
-    fkeys = [discover(fkey, t, parent_measure=Record(records))
+    ordering = {str(c): i for i, c in enumerate(c for c in t.columns.keys())}
+    record = list(_process_columns(t.columns))
+    fkeys = [discover(fkey, t, parent_measure=Record(record))
              for fkey in t.foreign_keys]
     for name, column in merge(*fkeys).items():
-        records[ordering[name]] = (name, column)
-    return var * Record(records)
+        record[ordering[name]] = (name, column)
+    return var * Record(record)
+
+
+def _process_columns(columns):
+    """Process the dshapes of the columns of a table.
+
+    Parameters
+    ----------
+    columns : iterable[column]
+        The columns to process.
+
+    Yields
+    ------
+    record_entry : tuple[str, dshape]
+        A record entry containing the name and type of each column.
+    """
+    for col in columns:
+        (name, dtype), = discover(col).fields
+        yield str(name), dtype
 
 
 @memoize
--- a/odo/backends/tests/test_sql.py
+++ b/odo/backends/tests/test_sql.py
@@ -9,6 +9,7 @@ from functools import partial
 
 import datashape
 from datashape import discover, dshape, float32, float64
+from datashape.util.testing import assert_dshape_equal
 import numpy as np
 import pandas as pd
 import sqlalchemy as sa
@@ -18,6 +19,8 @@ from odo.backends.sql import (
     dshape_to_table, create_from_datashape, dshape_to_alchemy
 )
 from odo.utils import tmpfile, raises
+
+from six import string_types
 
 
 def test_resource():
@@ -86,10 +89,10 @@ def test_discover():
                  sa.Column('name', sa.String),
                  sa.Column('amount', sa.Integer),
                  sa.Column('timestamp', sa.DateTime, primary_key=True))
-
-    assert discover(s) == \
-        dshape('var * {name: ?string, amount: ?int32, timestamp: datetime}')
-
+    ds = dshape('var * {name: ?string, amount: ?int32, timestamp: datetime}')
+    assert_dshape_equal(discover(s), ds)
+    for name in ds.measure.names:
+        assert isinstance(name, string_types)
 
 def test_discover_numeric_column():
     assert discover(sa.String()) == datashape.string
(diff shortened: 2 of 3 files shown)


llllllllll:
Thanks for the PR; I will look at this when I get into the office

odo/backends/sql.py
```diff
@@ -199,8 +199,10 @@ def discover_sqlalchemy_column(c):
 
 @discover.register(sa.sql.FromClause)
 def discover_sqlalchemy_selectable(t):
-    ordering = dict((c, i) for i, c in enumerate(c for c in t.columns.keys()))
+    ordering = dict((str(c), i)
```
llllllllll: we no longer support py2.6 so we can make this a dict comprehension.

llllllllll: I think we want `c.name` here instead of `str(c)`

odo/backends/sql.py
```diff
@@ -199,8 +199,10 @@ def discover_sqlalchemy_column(c):
...
+                    for i, c in enumerate(c for c in t.columns.keys()))
     records = list(sum([discover(c).parameters[0] for c in t.columns], ()))
+    records = [(str(r[0]), r[1]) for r in records]
```
llllllllll: The original expression is a little confusing, and iterating over the sequence again is a bit unclear. It took me a while to walk through the intent of this code, mostly because of what already existed though. Do you think we could pull this into another helper function? something like:

```python
def _process_columns(columns):
    """Process the dshapes of the columns of a table.

    Parameters
    ----------
    columns : iterable[column]
        The columns to process.

    Yields
    ------
    record_entry : tuple[str, dshape]
        A record entry containing the name and type of each column.
    """
    for col in columns:
        (name, dtype), = discover(col).fields
        yield str(name), dtype
```
here fields replaces parameters[0] because we know this should be a record type.

We can then replace this call site with something like:

```python
record = list(_process_columns(t.columns))
```

Also, we should probably call this `record` instead of `records` because it is the types for a single record and thus not plural.

I think this expression would be more clear but I am happy to hear what you think.

odo/backends/tests/test_sql.py
```diff
@@ -86,10 +88,10 @@ def test_discover():
                  sa.Column('name', sa.String),
                  sa.Column('amount', sa.Integer),
                  sa.Column('timestamp', sa.DateTime, primary_key=True))
-
-    assert discover(s) == \
-        dshape('var * {name: ?string, amount: ?int32, timestamp: datetime}')
-
+    ds = dshape('var * {name: ?string, amount: ?int32, timestamp: datetime}')
+    assert discover(s) == ds
```
llllllllll: since this test was added we have added `datashape.util.assert_dshape_equal`. This function provides much more helpful error messages in the failure case, for example, imagine the following two moderatly complex dshapes:

```python
a = dshape('var * {a: {b: ?int32, c: ?float64}, d: string[10]}')
b = dshape('var * {a: {b: ?int32, c: ?float32}, d: string[10]}')
```

These are slightly different, with assert a == b we get:

```
    def test_ex():
        a = dshape('var * {a: {b: ?int32, c: ?float64}, d: string[10]}')
        b = dshape('var * {a: {b: ?int32, c: ?float32}, d: string[10]}')
>       assert a == b
E       assert dshape("var *... string[10]}") == dshape("var * ... string[10]}")
E         Use -v to get the full diff

```

with `assert_dshape_equal` we get the path right to the error:
```
E           AssertionError: ctype names do not match: 'float64' != 'float32'
E           path: _.measure['a']['c'].ty.name
```

We can also pass a bunch of special arguments to change certain comparisons like ignoring the order of records or ignoring fixed lengths for strings.

odo/backends/sql.py
thequackdaddy: `sqlalchemy.sql.elements.quoted_name` does not have an attribute of `name`...

thequackdaddy: there is a function `__str__()`

I dunno... this seems longer... I'd prefer sticking with `str(c)`

odo/backends/sql.py
thequackdaddy: This is much more elegant, I agree. I've done some copypasta.

odo/backends/tests/test_sql.py
thequackdaddy: Added this test here.

thequackdaddy:
@llllllllll Ok. I've updated code and it's passing the unittests. 

I can squash/rebase, if that's helpful. Can I also bother you to talk me through #430. I don't think we want to commit that, but I'd like to make it so I can use MSSQL server with the blaze ecosystem. I'm unsure of how you would want these changes organized. 

Thanks.

llllllllll:
I will read through #430 now, thanks. Also, those are lowercase `L` characters, not pipes ;)
If you could squash and use `BUG` instead of `FIX` that would be awesome.

thequackdaddy:
Squashed.

llllllllll:
Oh, sorry but I forgot one thing: can you add an entry here: https://github.com/blaze/odo/blob/master/docs/source/whatsnew/0.5.0.txt#L36 which gives a short description of this fix and ends with a link to this issue like:

```rst
(:issue:`433`).
```

We use these file to generate the release notes. Other than that I will merge on pass

thequackdaddy:
Done.

llllllllll:
Thank you

PR #434, opened 2016-03-19
ENH: Indexing on chunked pandas

thequackdaddy:
Hello,

This makes the index work when you have an iterator go through a pandas dataframe. Otherwise, the index constantly repeats. A new test demonstrates this functionality. 

Thanks for you consideration.

### DIFF
--- a/odo/convert.py
+++ b/odo/convert.py
@@ -3,7 +3,7 @@ from __future__ import absolute_import, division, print_function
 import numpy as np
 import pandas as pd
 from datashape.predicates import isscalar
-from toolz import concat, partition_all
+from toolz import concat, partition_all, compose
 from collections import Iterator, Iterable
 import datashape
 from datashape import discover
@@ -206,6 +206,12 @@ def iterator_to_numpy_chunks(seq, chunksize=1024, **kwargs):
 @convert.register(chunks(pd.DataFrame), Iterator, cost=10.0)
 def iterator_to_DataFrame_chunks(seq, chunksize=1024, **kwargs):
     seq2 = partition_all(chunksize, seq)
+
+    if kwargs.get('add_index'):
+        mkindex = _add_index
+    else:
+        mkindex = _ignore_index
+
     try:
         first, rest = next(seq2), seq2
     except StopIteration:
@@ -213,12 +219,28 @@ def iterator_to_DataFrame_chunks(seq, chunksize=1024, **kwargs):
             yield convert(pd.DataFrame, [], **kwargs)
     else:
         df = convert(pd.DataFrame, first, **kwargs)
+        df1, n1 = mkindex(df, 0)
 
         def _():
-            yield df
+            n = n1
+            yield df1
             for i in rest:
-                yield convert(pd.DataFrame, i, **kwargs)
+                df = convert(pd.DataFrame, i, **kwargs)
+                df, n = mkindex(df, n)
+                yield df
     return chunks(pd.DataFrame)(_)
+
+
+def _ignore_index(df, start):
+    return df, start
+
+
+def _add_index(df, start, _idx_type=getattr(pd, 'RangeIndex',
+                                            compose(pd.Index, np.arange))):
+    stop = start + len(df)
+    idx = _idx_type(start=start, stop=stop)
+    df.index = idx
+    return df, stop
 
 
 @convert.register(tuple, np.record)
--- a/odo/tests/test_convert.py
+++ b/odo/tests/test_convert.py
@@ -3,7 +3,8 @@ from __future__ import absolute_import, division, print_function
 import pytest
 from odo.convert import (convert, list_to_numpy, iterator_to_numpy_chunks,
                          dataframe_to_chunks_dataframe, numpy_to_chunks_numpy,
-                         chunks_dataframe_to_dataframe)
+                         chunks_dataframe_to_dataframe,
+                         iterator_to_DataFrame_chunks)
 from odo.chunks import chunks
 from datashape import discover, dshape
 from collections import Iterator
@@ -228,7 +229,17 @@ def test_pandas_and_chunks_pandas():
     assert len(list(c)) == 2
 
     df2 = chunks_dataframe_to_dataframe(c)
-    assert str(df2) == str(df)
+    tm.assert_frame_equal(df, df2)
+
+
+def test_iterator_to_DataFrame_chunks():
+    data = ((0, 1), (2, 3), (4, 5), (6, 7))
+    df1 = pd.DataFrame(list(data))
+    df2 = iterator_to_DataFrame_chunks(data, chunksize=2, add_index=True)
+    df2 = pd.concat(df2, axis=0)
+    tm.assert_frame_equal(df1, df2)
+    df2 = convert(pd.DataFrame, data, chunksize=2, add_index=True)
+    tm.assert_almost_equal(df1, df2)
 
 
 def test_recarray():
--- a/docs/source/whatsnew/0.5.0.txt
+++ b/docs/source/whatsnew/0.5.0.txt
@@ -7,7 +7,9 @@ Release |version|
 New Features
 ------------
 
-None
+* For conversions that go through chunked pandas DataFrames, adding the
+  keyword `add_index=True` will append a unique index column onto the data.
+  Otherwise, the index will likely repeat each chunk.
 
 Experimental Features
 ---------------------


odo/tests/test_convert.py
```diff
@@ -228,7 +229,15 @@ def test_pandas_and_chunks_pandas():
...
 
     df2 = chunks_dataframe_to_dataframe(c)
-    assert str(df2) == str(df)
+    tm.assert_frame_equal(df, df2)
+
+
+def test_iterator_to_DataFrame_chunks():
+    data = ((0, 1), (2, 3), (4, 5), (6, 7))
+    df1 = pd.DataFrame(list(data))
+    df2 = iterator_to_DataFrame_chunks(data, 2)
```
llllllllll: This should probably black box test the conversion by calling `odo` instead of the underlying function.

odo/convert.py
```diff
@@ -213,14 +213,24 @@ def _():
...
+                df = convert(pd.DataFrame, i, **kwargs)
+                df, n = _add_index(df, n)
+                yield df
     return chunks(pd.DataFrame)(_)
 
 
+def _add_index(df, start):
+    idx = pd.Index(np.arange(start, start + len(df)))
+    df.index = idx
+    return df, start + len(df)
```
llllllllll: no need to do this computation twice, we can store it at the top in a local `stop`

odo/convert.py
```diff
@@ -213,14 +213,24 @@ def _():
...
             for i in rest:
-                yield convert(pd.DataFrame, i, **kwargs)
+                df = convert(pd.DataFrame, i, **kwargs)
+                df, n = _add_index(df, n)
+                yield df
     return chunks(pd.DataFrame)(_)
 
 
+def _add_index(df, start):
+    idx = pd.Index(np.arange(start, start + len(df)))
```
llllllllll: I feel like we should use a `RangeIndex` here for newer versions of pandas. We can conditionally define some `mkindex` function and maybe store it as a default arg to this `add_index` so that we don't pay a perf penalty for this.

llllllllll:
Blaze ignores the index on dataframes so having a non-unique index is not a problem. What is the performance overhead of adding this? I imagine that using a range index will make this very fast so I am not too worried I just want to make sure we know what the impact is. Thank you for this feature, could you please update the `whatsnew`, maybe as an API change.

thequackdaddy:
The point of this feature is that I was trying to get a big MSSQL table into HDFStore and I noticed that the index was repeating. That was inconvenient for me... made slicing particular records hard.

llllllllll:
I see, I think this is a more correct way to do this, do you think we should name the dataframe with the chunk number also or is this just overkill?

odo/tests/test_convert.py
thequackdaddy: I added a black box, but I couldn't (in about 30 seconds of trying) convince myself that this would get all the keywords right, so I did it both ways.

odo/convert.py
thequackdaddy: Fixed.

odo/convert.py
thequackdaddy: Fixed.

thequackdaddy:
Ok, I added a add_index keyword (easy to change... not sure if that's the best term... maybe add_index_pandas?). We check for it and assign the function `mkindex` based on the outcome. 

I'm not sure what you mean "name the dataframe"... Not sure how to order parts of an iterator...

thequackdaddy:
Oh, and I'll edit the whatsnew once we decide on the syntax for all this. I can squash then too.

odo/convert.py
```diff
@@ -206,6 +206,15 @@ def _():
 @convert.register(chunks(pd.DataFrame), Iterator, cost=10.0)
 def iterator_to_DataFrame_chunks(seq, chunksize=1024, **kwargs):
     seq2 = partition_all(chunksize, seq)
+
+    try:
+        if kwargs['add_index'] is True:
```
llllllllll: we should just check the truthiness of the value instead of doing pointer comparisons against 'true'.

Also, to remove the try/except we can just use `dict.get` like: `if kwargs.get('add_index'):`. This function does does a lookup and returns some value if the element is not in the dict, defaulting to None (which acts as false)

odo/convert.py
```diff
@@ -213,14 +222,29 @@ def _():
...
     return chunks(pd.DataFrame)(_)
 
 
+def _ignore_index(df, start):
+    return df, start
+
+
+def _add_index(df, start):
+    stop = start + len(df)
+    idx = pd.RangeIndex(start=start, stop=stop)
```
llllllllll: range index is added in pandas 17, unfortunatly some people still need to support pandas 16, could we conditionally define this based on the prescense of `RangeIndex`?

odo/convert.py
```diff
@@ -213,14 +222,29 @@ def _():
...
+            yield df1
             for i in rest:
-                yield convert(pd.DataFrame, i, **kwargs)
+                df = convert(pd.DataFrame, i, **kwargs)
+                df, n = mkindex(df, n)
+                yield df
     return chunks(pd.DataFrame)(_)
 
 
+def _ignore_index(df, start):
```
llllllllll: +1 on using an identity function to keep the rest of the logic simple

llllllllll:
Side note: you may use my actual name in commits instead of my github username. Sometimes the username is hard to read

thequackdaddy:
@llllllllll 

I love your username too much. I refuse to use your real name. I can settle for "barcode".

llllllllll:
Haha, okay, you can keep using the username.

kwmsmith:
It's a good thing @llllllllll didn't mix in `1` and `|` characters in :)

thequackdaddy:
Hello,

@kwmsmith Was this closed by accident? (I think the "Close and Comment" and "Comment" buttons are perilously close to one another.) I programmed the latest comments from @llllllllll just now. 

Happy to make a new PR if we can't re-open. I guess I should squash too.

llllllllll:
ping @kwmsmith, why did you close this, or was this just an accident?

kwmsmith:
> ping @kwmsmith, why did you close this, or was this just an accident?

My bad, apologies; thanks for reopening.

odo/convert.py
```diff
@@ -213,14 +219,32 @@ def _():
...
+    return df, start
+
+
+def _add_index(df, start):
+    stop = start + len(df)
+    if hasattr(pd, 'RangeIndex'):
```
llllllllll: I was thinking we could  pull this into a default argument to save us from rechecking at runtime, like: 

```python
def _add_index(df, start, _idx_type=getattr(pd, 'RangeIndex', pd.Index)):
```

this way we only do the lookup once

llllllllll:
Just one comment about where to check the index type, then add the whatsnew. Thank you again for this!

odo/convert.py
thequackdaddy: Sorry I'm a little confused here. I figured I could do this...

    def _add_index(df, start, _idx_type=getattr(pd, 'RangeIndex', np.arange)):
        stop = start + len(df)
        idx = pd.Index(_idx_type(start=start, stop=stop))
        df.index = idx
        return df, stop

But if we do `_idx_type=getattr(pd, 'RangeIndex', pd.Index))` I think we'd need to check whether RangeIndex successfully imported. The arguments to RangeIndex and Index appear to be quite different... I'm able to avoid an `if` statement with this.

llllllllll: ah, I forgot that these are slightly different, the case could be `toolz.compose(pd.Index, np.arange)`

odo/convert.py
```diff
@@ -213,14 +219,30 @@ def _():
...
 
 
+def _ignore_index(df, start):
+    return df, start
+
+
+def _add_index(df, start, _idx_type=getattr(pd, 'RangeIndex',
+                                            compose(pd.Index, np.arange))):
+    stop = start + len(df)
+    idx = pd.Index(_idx_type(start=start, stop=stop))
```
llllllllll: I don't think we still need to wrap this in Index because the result of `_idx_type` should be an index. Other that this lgtm

thequackdaddy:
Cool thanks. I squashed it too.

odo/convert.py
thequackdaddy: Oh geez. You are right.

thequackdaddy:
Ok. Squashed again. Hopefully this is good.

llllllllll:
Looks great, sorry about all the back and forth, merging on pass.

thequackdaddy:
Not a problem at all. You're comments are super helpful.

PR #439, opened 2016-04-04
When downloading a URL resource, ensure we read everything.

kwmsmith:


odo/backends/tests/test_url.py
```diff
@@ -34,6 +34,19 @@ def test_url_resource():
     assert isinstance(csv, URL(CSV))
 
 
+def test_small_chunk_size():
+    url_csv = resource(iris_url)
+    normal = convert(Temp(CSV), url_csv)
+    url_csv.chunk_size = 1
```
llllllllll: can we pass this as a kwarg to `resource` or is it expected to mutate the object after creation?

kwmsmith:
@llllllllll any comments on this one?  This is blocking us, so timely feedback is appreciated, thanks!

llllllllll:
looking now

odo/backends/url.py
```diff
@@ -133,7 +133,9 @@ def append_urlX_to_X(target, source, **kwargs):
         with open(target.path, 'wb') as fp:
             for chunk in iter(curry(r.read, chunk_size), ''):
                 fp.write(chunk)
-                return target
+                if not chunk:
```
llllllllll: Do we ever hit this branch? The second argument to iter here should prevent us from entering the body when we read an empty string, for example:

```python
In [13]: def f(it=iter(['a', 'b', 'c', ''])):
   ....:     return next(it)
   ....: 

In [14]: tuple(iter(f, ''))
Out[14]: ('a', 'b', 'c')
```

One thing we may want to do though is change this loop to break when `len(chunk) < chunk_size` because this will avoid that last empty read, but idk if that is worth it.

kwmsmith: We definitely hit this branch, without it, it's an infinite loop :)

The issue is that `http.client.HTTPResponse.read` returns `b''` forever when it's finished, so we have to explicitly test for this case and break.

We could check for `len(chunk) < chunk_size`, but this code here seems cleaner to me.

I'll add a comment explaining `HTTPResponse.read`'s behavior here.

llllllllll: sorry, I understand that it is returning an empty string, but maybe that means our `iter` is written incorrectly and should be `iter(curry(...), b'')`. This is what two argument iter is supposed to solve.

kwmsmith: Of course, I see what you mean now.  Fixing it to `iter(curry(...), b'')` fixes the issue, and we can remove the test and break.  Thanks!

kwmsmith:
Thanks @llllllllll for the feedback.

PR #448, opened 2016-05-10
Don't validate on get_bucket

jseabold:
Similar PR and rationale here https://github.com/conda/conda/pull/2126


llllllllll:
lgtm

PR #449, opened 2016-05-19
Greedy Retry

llllllllll:
When an edge fails by raising NotImplementedError we will not compute
the cost from the current object to the target and compare it to the
cost of converting from the source to the target. We will then continue
with the cheaper route.

Example:

Let A, B, C, and D be distinct types.
Imagine we have the following routes defined:

A->B cost=100.0
B->C cost=1.0
B->D cost=3.0
C->D cost=1.0

Now we will call convert(D, A()). The initial route selected will be:

A->B, B->C, C->D with a total cost of 102.00.

Assume that we have finished the A->B conversion and then the B-C edge
fails. Currently we would recalculate the route from A to D omitting the
B-C edge which leaves us with A->B, B->D with a total cost of 103.0
This means that we need to duplicate the 100.0 cost to recompute
A->B. This path makes convert consider the route listed above, but also
looks for routes from B to D. We will then see the route B->D with a
cost of 3.0. Because 3.0 is less than the alternative best route of cost
103.0 we will select this new route instead.

side note:

our releases page was _very_ out of date


llllllllll:
ping @kwmsmith @sandhujasmine

llllllllll:
bump

llllllllll:
bump, I think I am going to merge this soon if there are no comments.

kwmsmith:
@llllllllll LGTM.

llllllllll:
The py2 builds are hanging at test discovery. I will investigate this soon. Assuming nothing radical changes based on that I will merge when that is fixed.

PR #450, opened 2016-05-22
Allow aws profiles

jseabold:
Allow `profile_name` for getting s3 client. This allows respecting of things like `~/.aws/credentials` so you can have multiple profiles and don't have to munge environment variables or parse the config file yourself.

I also caught another `get_bucket` without `validate=False`.

### DIFF
--- a/odo/backends/aws.py
+++ b/odo/backends/aws.py
@@ -27,8 +27,12 @@ from ..utils import tmpfile, ext, sample, filter_kwargs, copydoc
 @memoize
 def get_s3_connection(aws_access_key_id=None,
                       aws_secret_access_key=None,
-                      anon=False):
+                      anon=False, **kwargs):
     import boto
+
+    if 'profile_name' in kwargs:
+        return boto.connect_s3(profile_name=kwargs['profile_name'])
+
     cfg = boto.Config()
 
     if aws_access_key_id is None:
@@ -70,7 +74,8 @@ class _S3(object):
             self.s3 = s3
         else:
             self.s3 = get_s3_connection(aws_access_key_id=aws_access_key_id,
-                                        aws_secret_access_key=aws_secret_access_key)
+                                        aws_secret_access_key=aws_secret_access_key,
+                                        **kwargs)
         try:
             bucket = self.s3.get_bucket(self.bucket,
                                         **filter_kwargs(self.s3.get_bucket,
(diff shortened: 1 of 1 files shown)


odo/backends/aws.py
```diff
@@ -70,9 +73,11 @@ def __init__(self, uri, s3=None, aws_access_key_id=None,
             self.s3 = s3
         else:
             self.s3 = get_s3_connection(aws_access_key_id=aws_access_key_id,
-                                        aws_secret_access_key=aws_secret_access_key)
+                                        aws_secret_access_key=aws_secret_access_key,
+                                        **kwargs)
         try:
             bucket = self.s3.get_bucket(self.bucket,
+                                        validate=False,
```
kwmsmith: This shouldn't be necessary with the `**filter_kwargs(...)` unpacking on the next line -- if the user passes in `validate=False` to the constructor, it should make it through via `filter_kwargs()`.

That said, it may be necessary to do some work to ensure top-level `kwargs` make it all the way down to this function.

I wouldn't want to hardcode `validate=False` from skimming this: http://docs.pythonboto.org/en/latest/s3_tut.html#accessing-a-bucket

kwmsmith:
> I also caught another get_bucket without validate=False.

OK, sorry, I just re-read this; what's the rationale here for hardcoding `validate=False`?  What about the warning here?

http://docs.pythonboto.org/en/latest/s3_tut.html#accessing-a-bucket

Isn't it possible to pass in a `validate=False` at the top-level `odo` call and have this do what you want?

jseabold:
The practical reason for `validate=False` is that there's no way to use the library if you don't have list permissions on the bucket but rather have permissions only on some underlying prefixes.

We had to go through and add these to many of the conda / dask projects. You can see some of the discussion [here](https://github.com/conda/conda/pull/2126#issuecomment-189312777)

jseabold:
Let me try to remember why it didn't work at the top-level. FWIW, I find it a more sensible default, and the boto people changed to this in boto3. I'll see about not hard-coding it.

jseabold:
Rebased and got rid of the hard-coded validation. Will see if I can just get by with sticking a `validate=False` everywhere I need to use s3.

jseabold:
I guess this is moot now since #451 got merged.

PR #451, opened 2016-05-22
Support Redshift UNLOADs

### LINKED ISSUE #447: Support for redshift UNLOAD to S3
Parking this here, as it doesn't look like it's supported. 

Redshift will unload potentially multiple CSVs to an s3 [prefix](http://docs.aws.amazon.com/redshift/latest/dg/r_UNLOAD.html). This is supported by the [redshift dialect](http://sqlalchemy-redshift.readthedocs.io/en/latest/commands.html#sqlalchemy_redshift.commands.UnloadFromSelect) just like the CopyCommand that's used.

Happy to look into this, if you have any pointers.

I also asked [here](https://github.com/blaze/odo/issues/34#issuecomment-215593436) to connect the conversations.

jseabold:
I'm sure this will need a once over. I don't have a handle on all of the internals yet. Questions I have
- Is a Directory(S3) the right resource type for an s3 prefix (an s3 uri that ends in a forward slash)?
- What should be tested?
- Is it customary to leave out explicit keyword arguments in things like redshift_to_s3_bucket? This rubs me the wrong way.
- Is append the right and only dispatcher I need?

Will need to rebase on #450.

Closes #447.

### DIFF
--- a/odo/backends/sql.py
+++ b/odo/backends/sql.py
@@ -35,6 +35,7 @@ from toolz import (partition_all, keyfilter, valfilter, identity, concat,
 from toolz.curried import pluck, map
 
 from ..compatibility import unicode, StringIO
+from ..directory import Directory
 from ..utils import (
     keywords,
     ignoring,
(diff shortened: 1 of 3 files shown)


kwmsmith:
@jseabold apologies for the extreme lag in looking in this.  Thanks for the PR.

Assuming this is still relevant to you, could you rebase / merge in master?  There's been some work around updating the sqlachemy-redshift imports that I want to make sure are consistent with this PR.

jseabold:
Rebased on master. I think it would be nice if we can find a way to get this in. I've written and am using my own redshift connector for dask but perhaps code sharing is better here.

Some of the fundamental questions asked above are still outstanding.

odo/backends/sql.py
```diff
@@ -842,6 +843,34 @@ def compile_copy_to_csv_sqlite(element, compiler, **kwargs):
...
+    from odo.backends.aws import S3, get_s3_connection
+except ImportError:
+    pass
+else:
+    @resource.register('s3://.*/$')
+    def resource_s3_prefix(uri, **kwargs):
+        return Directory(S3)(uri, **kwargs)
+
+    @append.register(Directory(S3), sa.Table)
+    def redshit_to_s3_bucket(bucket, selectable, dshape=None, bind=None,
```
davidthewatson: Was the mis-spelling of redshift here intentional?

davidthewatson:
Is this going to be merged into master? I need this in blaze.

odo/backends/sql.py
llllllllll: I assume this was a mistake, we can fix this post-merge.

llllllllll:
I don't think anything is still blocking this. I am going to merge. If there is a problem we can revert.

PR #466, opened 2016-08-02
Pass kwargs to S3 append_text_to_s3 functions

### LINKED ISSUE #166: Write to S3 using server-side encryption
As a developer, I want to copy data to S3 such that the resulting objects are encrypted, so that I can use Odo with sensitive data.

boto's set_contents_from_filename() can take an optional parameter encrypt=True, so there needs to be a way to pass that through the call chain such as append_text_to_s3().

Acceptance criteria:
Save data to S3 in a bucket such that the resulting object is encrypted with AES-256. One way to test this is that a bucket with the following policy, will not allow unencrypted puts:
```
{
   "Version":"2012-10-17",
   "Id":"PutObjPolicy",
   "Statement":[{
         "Sid":"DenyUnEncryptedObjectUploads",
         "Effect":"Deny",
         "Principal":"*",
         "Action":"s3:PutObject",
         "Resource":"arn:aws:s3:::mybucket/*",
         "Condition":{
            "StringNotEquals":{
               "s3:x-amz-server-side-encryption":"AES256"
            }
         }
      }
   ]
}
```

dan-coates:
- This allows encrypt_key=True to be passed to the relevant boto functions,
  which is required if the S3 bucket requires server-side encrpytion.

I wrote two new tests to test this functionality, but to test it properly, you need an S3 bucket with a policy that will only accept puts with server-side encryption. So as not to break CI, I configured this so that the tests will run if the ODO_S3_ENCRYPTION_BUCKET environment variable is set. This environment variable should contain the name of a bucket that has this policy.

I tested this locally and the tests fail without this patch and pass with the patch. If there's a solid way to always have a bucket with this policy on hand for the tests, it can be made non-optional, otherwise I'm comfortable that this resolves #166.

### DIFF
--- a/odo/backends/tests/test_s3.py
+++ b/odo/backends/tests/test_s3.py
@@ -93,34 +93,34 @@ test_bucket_name = 'into-redshift-csvs'
 _tmps = ('tmp%d' % i for i in itertools.count())
 
 
-def test_s3_encrypted_upload():
+@pytest.fixture
+def s3_encryption_bucket():
     test_bucket = os.getenv('ODO_S3_ENCRYPTION_BUCKET')
     if not test_bucket:
         pytest.skip('No bucket defined that requires server-side encryption')
+    return test_bucket
 
+
+def test_s3_encrypted_upload(s3_encryption_bucket):
     s3_connection = boto.connect_s3()
 
     df = tm.makeMixedDataFrame()
     with tmpfile('.csv') as fn:
         df.to_csv(fn, index=False)
-        s3_uri = 's3://{bucket}/{fn}'.format(bucket=test_bucket, fn=os.path.basename(fn))
+        s3_uri = 's3://{bucket}/{fn}'.format(bucket=s3_encryption_bucket, fn=os.path.basename(fn))
         odo(fn, s3_uri, s3=s3_connection, encrypt_key=True)
         result = odo(s3_uri, pd.DataFrame, s3=s3_connection)
 
     tm.assert_frame_equal(df, result)
 
 
-def test_s3_encrypted_multipart_upload():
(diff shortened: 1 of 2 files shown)


kwmsmith:
@octophat thanks for the work and the writeup.

We're seeing these travis-ci failures everywhere now, and we're having some trouble getting to the bottom of what's causing them; once that's resolved we'll re-run travis on this PR and merge if they succeed.

kwmsmith:
@dan-coates could you merge in `master` to this PR?  A recent PR fixed the travis-ci config / setup error.

dan-coates:
@kwmsmith I did a rebase rather than a merge to avoid more merge commits, which perhaps was a bad idea, but I think it should be all good now.

odo/backends/tests/test_s3.py
```diff
@@ -93,6 +93,40 @@ def conn():
 _tmps = ('tmp%d' % i for i in itertools.count())
 
 
+def test_s3_encrypted_upload():
+    test_bucket = os.getenv('ODO_S3_ENCRYPTION_BUCKET')
+    if not test_bucket:
+        pytest.skip('No bucket defined that requires server-side encryption')
```
kwmsmith: Please refactor this block into a top-level test fixture, something like

```python
@pytest.fixture
def s3_encryption_bucket():
    test_bucket = os.getenv(...)
    # pytest.skip logic if test_bucket is None
    return test_bucket
```

and declare the fixture as an arg to `test_s3_encrypted_upload(s3_encription_bucket)` and  `test_s3_encrypted_multipart_upload`

dan-coates:
Thanks for the feedback - I have much to learn on pytest to break me out of my nose comfort zone.

PR #467, opened 2016-08-03
Issue #465: Fix datashape discovery if dshape=None.

skrah:
@kwmsmith This fixes #465.  If you have time to merge this, it might be easier to finish work on https://github.com/blaze/blaze/pull/1559 .

PR #468, opened 2016-08-03
Remove build directory in the travis builds.

skrah:
The build failures occur whenever there's a leftover build directory.  Let's see if this fixes it.


skrah:
Too bad. The build directory was the only way I could reproduce the failures from blaze/odo#467 on my machine.

skrah:
Closing, since this wasn't the cause on Travis CI.

PR #470, opened 2016-08-04
Use boto3 from conda and disable pip --upgrade.

skrah:


skrah:
The Travis CI failures occur when pip pulls in boto3-1.4.0-py2.py3-none-any.whl.  Using boto3 from conda "fixes" the issue.

skrah:
@kwmsmith I think this could be a reasonable fix until someone more familiar with boto3 tracks down the boto3 version issues.

skrah:
N.B. I cannot reproduce this at home either with boto3-1.4.0, so it will remain a CI mystery. :)

kwmsmith:
@skrah thank you for getting to the bottom of this!

PR #471, opened 2016-08-04
Build failures minimal patch

skrah:
Test if dropping ```--upgrade``` is sufficient.


skrah:
This doesn't work either: #470 is the winner.

PR #473, opened 2016-08-09
S3(x) to x directly gets object from S3

### LINKED ISSUE #472: S3(CSV) to CSV routes through DataFrames unnecessarily
I’m testing out odo’s s3 functionality with a round trip of local CSV -> S3 CSV -> local CSV. The first leg of this round trip seems to work fine, but on the return trip, it’s not simply getting the file as I would expect. Instead, the CSV ends up with numeric fields clearly manipulated (e.g. 39.00 is changed to 39.0).

Looking into what’s happening here, it looks like the S3(CSV) to CSV path is happening through a conversion to a Temp(CSV) first (hitting [`s3_text_to_text`](https://github.com/blaze/odo/blob/0.5.0/odo/backends/aws.py#L213)), which then results in hitting [`append_object_to_csv`](https://github.com/blaze/odo/blob/0.5.0/odo/backends/csv.py#L201), which will convert the CSV into and out of a DataFrame, because there is no explicit Temp(CSV) to CSV call defined.

I’m a bit confused as to why this doesn’t just call the S3 object’s `get_contents_to_filename` method, as [`s3_text_to_temp_text`](https://github.com/blaze/odo/blob/0.5.0/odo/backends/aws.py#L203) does. Perhaps it was assumed that there would be a Temp(x) -> x path defined everywhere and the move from S3 to temp would just take care of everything. Since there’s not, I think we should just get the file directly to the output path given.

Since this is blocking me right now, I’ll work on a fix and submit a PR if/when I get it working. Just wanted to put this issue out there for any discussion on the expected behavior.

dan-coates:
It used to route through a Temp(*) object unnecessarily.
Also removed an unneeded import from the S3 tests.

Resolves #472.

### DIFF
--- a/odo/backends/aws.py
+++ b/odo/backends/aws.py
@@ -211,7 +211,8 @@ def s3_text_to_temp_text(s3_object, **kwargs):
 @append.register(JSONLines, S3(JSONLines))
 @append.register(TextFile, S3(TextFile))
 def s3_text_to_text(data, s3_object, **kwargs):
-    return append(data, convert(Temp(s3_object.subtype), s3_object, **kwargs), **kwargs)
+    s3_object.object.get_contents_to_filename(data.path)
+    return data
 
 
 @append.register((S3(CSV), Temp(S3(CSV))), (S3(CSV), Temp(S3(CSV))))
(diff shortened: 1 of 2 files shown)


kwmsmith:
@dan-coates I merged a workaround to handle the paramiko hang in test_ssh -- please merge / rebase on master, thanks!

kwmsmith:
Thanks @dan-coates.

PR #476, opened 2016-08-16
adding dshape parameter to CSV

### LINKED ISSUE #461: potential mismatch of data type for csv column
Due to the way of discovering the datashape of a csv file:

def discover_csv(c, nrows=1000, **kwargs):

by default, it only checks the first 1000 rows, so it is possible (For example, in a column with sparse data -- mostly 0 with scattered float number) that later rows may contain data of a different type (which does not appear in the 1000 sampled rows) and cause dtype error when actually loading the data.

ywang007:
closes #461

### DIFF
--- a/odo/backends/tests/test_csv.py
+++ b/odo/backends/tests/test_csv.py
@@ -398,6 +398,13 @@ def test_discover_with_dotted_names():
     assert dshape == datashape.dshape('var * {"a.b": int64, "c.d": int64}')
     assert dshape.measure.names == [u'a.b', u'c.d']
 
+def test_discover_csv_with_fixed_dshape():
+    with filetext('name,val\nAlice,1\nBob,2') as fn:
+        ds = datashape.dshape('var * {name: string, val: float64}')
+        csv = CSV(fn, dshape=ds)
+        ds1 = discover(csv)
+        assert ds1 == ds
+
 
 try:
     unichr
--- a/odo/backends/csv_ok.py
+++ b/odo/backends/csv_ok.py
@@ -18,7 +18,7 @@ import pandas as pd
 import datashape
 
 from datashape import discover, Record, Option
-from datashape.predicates import isrecord, isdimension
+from datashape.predicates import isrecord
 from datashape.dispatch import dispatch
 
 from ..compatibility import unicode, PY2
@@ -140,8 +140,8 @@ class CSV(object):
         If the csv file has a header or not
     encoding : str (default utf-8)
         File encoding
-    dshape: datashape or string representation
-        used specified datashape
+    _dshape: datashape or string representation
+        pre-determined datashape
     kwargs : other...
         Various choices about dialect
     """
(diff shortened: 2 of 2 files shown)


odo/backends/tests/test_mysql.py
```diff
@@ -202,8 +202,8 @@ def test_sql_to_csv(sql, csv):
         csv = odo(sql, fn)
         assert odo(csv, list) == data
 
-        # explicitly test that we do NOT preserve the header here
-        assert discover(csv).measure.names != discover(sql).measure.names
+        # explicitly test that we do NOT preserve the header here ???
+        #assert discover(csv).measure.name != discover(sql).measure.name
```
ywang007: Is there a reason the header should NOT be preserved? Or was it just not preserved before because we kept rediscovering the datashape?

odo/backends/csv.py
```diff
@@ -140,18 +140,25 @@ class CSV(object):
...
+                 sniff_nbytes=10000, dshape=None, **kwargs):
         self.path = path
         self._has_header = has_header
         self.encoding = encoding or 'utf-8'
         self._kwargs = kwargs
         self._sniff_nbytes = sniff_nbytes
+        if dshape:
+            if isinstance(dshape, (str, unicode)):
+                dshape = datashape.dshape(dshape)
+            dshape = None if isdimension(dshape.subshape[0][0])  else dshape
```
kwmsmith: What is this logic for?  Don't you want to test if `isrecord(dshape)` and raise an exception if `False`?

If an invalid dshape is passed in we should raise an exception, not silently swallow it...

odo/backends/csv.py
```diff
@@ -330,6 +337,9 @@ def _():
 
 @discover.register(CSV)
 def discover_csv(c, nrows=1000, **kwargs):
+    if c._dshape:
+        return c._dshape
```
kwmsmith: Perhaps we could add an `ensure_consistent_dshape` default arg (default `False`); if set to `True`, then the `c._dshape` would be tested against the dshape of the `df` loaded below to ensure the user-specified dshape is compatible with the discovered dshape.

odo/backends/tests/test_mysql.py
kwmsmith: Why was this commented out?

odo/backends/tests/test_csv.py
```diff
@@ -398,6 +398,13 @@ def test_discover_with_dotted_names():
     assert dshape == datashape.dshape('var * {"a.b": int64, "c.d": int64}')
     assert dshape.measure.names == [u'a.b', u'c.d']
 
+def test_discover_csv_with_fixed_dshape():
+    with filetext('name,val\nAlice,1\nBob,2') as fn:
+        ds = datashape.dshape('var * {name: string, val: float64}')
+        csv = CSV(fn, dshape=ds)
+        ds1 = discover(csv)
+        assert ds1 == ds
```
kwmsmith: We should add a test that verifies that the passed-in datashape overrides the datashape when it isn't passed in.

Perhaps a CSV file like:

```
a,b
1,1.0
 , 
2,2.0
```

And an overridden dshape like `var * {a: ?int32, ?int64}`.

ywang007:
@kwmsmith, since we are adding the datashape parameter for a very specific purpose, that is to allow user force one when he's not satisfied with the discovered one, I figure it's better to use a different variable name, user_dshape, to make it more explicit, and avoid conflicts with any internal use of "dshape" parameter by odo. This will require us to update blaze doc accordingly.

PR #479, opened 2016-08-25
Handle low-precision Decimal data shapes

necaris:
Optional decimals with suitably low precision (i.e. `?decimal[precision<=2]` are translated as `float8` which is not a valid datashape (or NumPy datatype) -- we need to handle that case.


kwmsmith:
@necaris LGTM Can you add an entry to whatsnew?  I'll merge then.

necaris:
Will do

necaris:
@kwmsmith Added entry to whatsnew in 52b532d 😄

llllllllll:
I needed to move the whatsnew entry since we released 0.5.1. closing in favor of: #504, all of your commits are still there.

PR #481, opened 2016-09-19
BUG: fix globbed csv to chunked dataframe

llllllllll:
Fixes the test failure in blaze. This adds a test in odo to check for this.

### DIFF
--- a/odo/backends/csv.py
+++ b/odo/backends/csv.py
@@ -15,6 +15,7 @@ from toolz import concat, keyfilter, keymap, merge, valfilter
 
 import pandas as pd
 
+from dask.threaded import get as dsk_get
 import datashape
 
 from datashape import discover, Record, Option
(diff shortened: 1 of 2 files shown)


odo/backends/csv.py
```diff
@@ -366,8 +367,19 @@ def resource_glob(uri, **kwargs):
 @convert.register(chunks(pd.DataFrame), (chunks(CSV), chunks(Temp(CSV))),
                   cost=10.0)
 def convert_glob_of_csvs_to_chunks_of_dataframes(csvs, **kwargs):
```
mtydykov: Does this need a docstring?

odo/backends/csv.py
```diff
@@ -366,8 +367,19 @@ def resource_glob(uri, **kwargs):
...
-    data = [partial(convert, chunks(pd.DataFrame), csv, **kwargs) for csv in csvs]
-    return chunks(pd.DataFrame)(data)
+    f = partial(convert, chunks(pd.DataFrame), **kwargs)
+
+    def df_gen():
+        dsk = {}
+        p = []
+        for n, csv_ in enumerate(csvs):
```
mtydykov: It seems like we're just building a dictionary that tells `dsk_get` to apply the same function to all csv objects. Why is that necessary?

odo/backends/csv.py
llllllllll: No, this is an internal function so it is fine. Also, the signature is implied because it is a `convert` dispatch

odo/backends/csv.py
llllllllll: `dask.threaded.get` will run these function calls concurrently.

mtydykov: Ok. Then it might be good to add a comment that explains why the loop is necessary.

PR #484, opened 2016-10-10
Change attr access based on dask bag API change.

kwmsmith:
`dask.bag.from_filenames` -> `dask.bag.from_text`


llllllllll:
We require a minimum version of dask that no longer knows about the old name.

PR #485, opened 2016-10-10
pytest skip fixup

kwmsmith:


llllllllll:
This was addressed on master already. If I missed something please reopen.

PR #487, opened 2016-10-13
Fix Postgresql CSV import for older versions.

giladarnold:
Postresql's COPY command syntax has changed from 8.x to 9.x. odo uses
the new syntax, which produces a syntax error when running against 8.x
and older. Below is a patch that probes the database version and applies
the correct syntax. It also fixes a (suspected) bug in how encoding is
inferred.


odo/backends/sql_csv.py
```diff
@@ -151,32 +151,55 @@ def compile_from_csv_postgres(element, compiler, **kwargs):
...
-            quotechar=element.quotechar,
-            escapechar=element.escapechar,
-            header=element.header,
-            encoding=element.encoding or element.bind(
-                'show client_encoding'
-            ).execute().scalar()
-        ),
-        **kwargs
-    )
+    postgres_version = element.bind.execute('select version()').scalar()
```
llllllllll: I am worried about the perf penalty we will pay to check this on every insert. What do you think about creating a `WeakKeyDictionary` from `bind` to version? This will let us save these checks across calls.

odo/backends/sql_csv.py
```diff
@@ -151,32 +151,55 @@ def compile_from_csv_postgres(element, compiler, **kwargs):
...
+              """.format(compiler.preparer.format_table(element.element))
+          ).bindparams(
+              path=os.path.abspath(element.csv.path),
+              delimiter=element.delimiter,
+              na_value=element.na_value,
+              quotechar=element.quotechar,
+              escapechar=element.escapechar,
+              header=element.header,
+              encoding=element.encoding or element.bind.execute(
+                  'show client_encoding'
```
llllllllll: We could probably also memoize the encoding if we are caching the version.

odo/backends/sql_csv.py
```diff
@@ -151,32 +151,55 @@ def compile_from_csv_postgres(element, compiler, **kwargs):
...
-            header=element.header,
-            encoding=element.encoding or element.bind(
-                'show client_encoding'
-            ).execute().scalar()
-        ),
-        **kwargs
-    )
+    postgres_version = element.bind.execute('select version()').scalar()
+    if int(postgres_version.split()[1].split('.')[0]) >= 9:
+      return compiler.process(
```
llllllllll: in order to make this easier to follow, what do you think about breaking these two calls into helper functions like `_process_pg_8_or_lower` or `process_pg_9_and_greater`?

PR #489, opened 2016-10-18
Update Odo documentation/code for sqlalchemy_redshift

RAbraham:
This request combines the work of two authors to update documentation and code to refer to `sqlalchemy_redshift`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,15 +56,12 @@ install:
 
   # Install various deps
   - conda uninstall toolz
-  - pip install -U toolz sas7bdat psycopg2 dill 'pymongo<3'
+  - pip install -U toolz sas7bdat psycopg2 dill 'pymongo<3' sqlalchemy-redshift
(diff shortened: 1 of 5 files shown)


RAbraham:
@branning I don't know if you get notifications from this PR, but in case you don't, Travis failed on a change made for the travis file commit.

branning:
It's because Spark 1.4 and 1.5 aren't available any more (they're very old by now). That's not something we changed. You'll notice that all recent Travis runs have the same problem. Not sure there's much we can do in the context of this PR. Odo's automated testing is basically completely broken.

llllllllll:
We changed how we setup travis so that we don't have this spark issue again. Instead of putting this in the conda call in the .travis file can you move this into a new install target like `etc/requirments_redshift.txt` and then add `redshift` to the list in `extras_require` in setup.py. You will also want to put the strictly equals version in `requirements_ci.txt` This is so that we call all reproduce the versions installed when running the automated tests.

PR #490, opened 2016-10-18
ENH: use postgres copy from std(in|out) for perf

llllllllll:
This is a pretty nice performance win. It halved the runtime of one of my data loading applications.

I also have a bunch of changes to get travis passing again.

### DIFF
--- a/odo/backends/csv.py
+++ b/odo/backends/csv.py
@@ -7,6 +7,7 @@ import gzip
 import bz2
 import uuid
 import csv
+from tempfile import NamedTemporaryFile
 
 from glob import glob
 from contextlib import contextmanager
@@ -78,11 +79,26 @@ def alias(key):
     return aliases.get(key, key)
 
 
+class ModeProxy(object):
+    """Gzipped files use int enums for mode so we want to proxy it to provide
+    The proper string version.
+    """
+    def __init__(self, file, mode):
+        self._file = file
+        self.mode = mode
+
+    def __getattr__(self, attr):
+        return getattr(self._file, attr)
+
+    def __iter__(self):
+        return iter(self._file)
+
+
 @contextmanager
 def open_file(path, *args, **kwargs):
     f = compressed_open.get(ext(path), open)(path, *args, **kwargs)
     try:
-        yield f
+        yield ModeProxy(f, kwargs.get('mode', 'r'))
     finally:
         f.close()
 
@@ -148,12 +164,13 @@ class CSV(object):
     canonical_extension = 'csv'
 
     def __init__(self, path, has_header=None, encoding='utf-8',
-                 sniff_nbytes=10000, **kwargs):
+                 sniff_nbytes=10000, buffer=None, **kwargs):
         self.path = path
         self._has_header = has_header
         self.encoding = encoding or 'utf-8'
         self._kwargs = kwargs
         self._sniff_nbytes = sniff_nbytes
+        self._buffer = buffer
 
     def _sniff_dialect(self, path):
         kwargs = self._kwargs
@@ -180,16 +197,32 @@ class CSV(object):
                                                            encoding=self.encoding)
         return self._has_header
 
+    def open(self, mode='rb', **kwargs):
+        buf = self._buffer
+        if buf is not None:
+            buf.seek(0)
+            return buf
+
+        return open_file(self.path, mode=mode, **kwargs)
+
 
 @sample.register(CSV)
 @contextmanager
 def sample_csv(csv, length=8192, **kwargs):
-    should_delete = not os.path.exists(csv.path)
-    if should_delete:
-        with open_file(csv.path, mode='wb'):
+    should_delete = False
+    if csv.path is None:
+        mgr = NamedTemporaryFile(suffix='.csv', mode='wb+')
+    elif not os.path.exists(csv.path):
+        with csv.open(mode='ab'):
+            # ensure the file gets created
             pass
+        mgr = csv.open(mode='rb')
+        should_delete = True
+    else:
+        mgr = csv.open(mode='rb')
+
     try:
-        with open_file(csv.path, mode='rb') as f:
+        with mgr as f:
             with tmpfile(csv.canonical_extension) as fn:
                 with open(fn, mode='wb') as tmpf:
                     tmpf.write(f.read(length))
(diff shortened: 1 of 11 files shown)


odo/backends/sql.py
```diff
@@ -795,12 +807,19 @@ def append_table_to_csv(csv, selectable, dshape=None, bind=None, **kwargs):
                        merge(csv.dialect, kwargs))
     stmt = CopyToCSV(
         selectable,
-        os.path.abspath(csv.path),
+        os.path.abspath(csv.path) if csv.path is not None else None,
```
kglowinski: I see where you have the `compiles` for `CopyToCSV` but not where path is used. What are the repercussions of the statement being executed without a path backing it?

odo/backends/sql_csv.py
```diff
@@ -85,22 +91,18 @@ def compile_from_csv_sqlite(element, compiler, **kwargs):
...
-        stdin=subprocess.PIPE
+        stdin=subprocess.PIPE,
     ).decode(sys.getfilesystemencoding())
     if stderr:
-        # TODO: this seems like a lot of rigamarole
-        try:
-            raise OSError(stderr)
-        except OSError as e:
-            raise sa.exc.DatabaseError(' '.join(cmd), [], e)
+        raise sa.exc.DatabaseError(' '.join(cmd), [], OSError(stderr))
```
kglowinski: Is the OSError no longer needed?

odo/backends/sql_csv.py
```diff
@@ -229,6 +230,101 @@ def append_csv_to_sql_table(tbl, csv, bind=None, **kwargs):
...
+    def _save_chunk(self, start_i, end_i):
+
+        data_index = self.data_index
+
+        # create the data for a chunk
+        slicer = slice(start_i, end_i)
+        for i in range(len(self.blocks)):
+            b = self.blocks[i]
+            na_rep = 'nan' if b.dtype.kind == 'f' else self.na_rep
+            d = b.to_native_types(slicer=slicer,
```
kglowinski: Can we format this so the first arg starts on the newline?

odo/backends/sql_csv.py
```diff
@@ -19,6 +22,8 @@
 
 from ..append import append
 from ..convert import convert
+from ..compatibility import StringIO
+from ..utils import literal_compile
 from .csv import CSV, infer_header
 from ..temp import Temp
 from .aws import S3
```
kglowinski: Unrelated, but this import appears to be unused, since there's a try/except for importation farther down.

odo/backends/sql_csv.py
```diff
@@ -229,6 +230,101 @@ def append_csv_to_sql_table(tbl, csv, bind=None, **kwargs):
...
+# The set of sql dialects which do not bounce dataframes to disk to append
+# to a sql table.
+DATAFRAME_TO_TABLE_IN_MEMORY_CSV = frozenset({
+    'postgresql',
+})
+
+# TODO: figure out how to dynamically detect this number, users may have
+# compiled their sqlite with a different max variable number but this is the
+# default.
+SQLITE_MAX_VARIABLE_NUMBER = 999
```
kglowinski: Will there be a problem if people have a different sqlite max variable number?

odo/backends/sql_csv.py
```diff
@@ -229,6 +230,101 @@ def append_csv_to_sql_table(tbl, csv, bind=None, **kwargs):
...
+        with bind.begin() as c:
+            with csv.open() as f:
+                c.connection.cursor().copy_expert(literal_compile(stmt), f)
+    else:
+        with bind.begin() as conn:
+            conn.execute(stmt)
     return tbl
+
+
+class NanIsNotNullFormatter(CSVFormatter):
```
kglowinski: I see the test for this, where NaN's should stay NaN because spite of this wrapper- is there a test for the opposite of this? Where Null should be inserted correctly if not float?

odo/backends/csv.py
```diff
@@ -224,16 +282,11 @@ def append_dataframe_to_csv(c, df, dshape=None, **kwargs):
             kwargs['encoding'] = encoding
         elif sys.version_info[0] == 2:
             kwargs['mode'] = 'ab' if sys.platform == 'win32' else 'at'
-        f = compressed_open[ext(c.path)](c.path, **kwargs)
-    else:
-        f = c.path
 
-    try:
+    with c.open(mode=kwargs.get('mode', 'a')) as f:
```
llllllllll: Which line should I match? I think this is clear that we are passing a default mode

odo/backends/csv.py
```diff
@@ -278,17 +331,14 @@ def _csv_to_dataframe(c, dshape=None, chunksize=None, **kwargs):
...
 
-    compression = kwargs.pop('compression',
-                             {'gz': 'gzip', 'bz2': 'bz2'}.get(ext(c.path)))
-
     # See read_csv docs for header for reasoning
     if names:
         try:
-            found_names = pd.read_csv(c.path, encoding=encoding,
-                                      compression=compression, nrows=1)
+            with c.open() as f:
```
llllllllll: it defaults to rb

odo/backends/sql.py
```diff
@@ -649,6 +655,13 @@ def drop(table, bind=None):
 
 @convert.register(pd.DataFrame, (sa.sql.Select, sa.sql.Selectable), cost=200.0)
 def select_or_selectable_to_frame(el, bind=None, **kwargs):
```
llllllllll: element

odo/backends/sql.py
llllllllll: It is used in the `compiles` dispatches. Currently the postgres version does not need a path.

odo/backends/sql_csv.py
llllllllll: If you have compiled a larger max variable number then this will just chunk more than needed, if it is smaller it wont work. I have not encountered anyone with a smaller max variable number in practice, but we could add this as an arg to append_dataframe_to_sql_table.

odo/backends/sql_csv.py
llllllllll: Added a test for datetime

kglowinski:
👍

gryBox:
Hi - Is there a release date for this issue? Am I understanding correctly that speeds of ~ 75 MBs  can achieved using this method?

Thanks in advance

llllllllll:
I am not sure when we want to cut a release, @kwmsmith do you have an ETA on the next release?

llllllllll:
Would you like to open a PR to fix this?

PR #496, opened 2016-11-02
BLD: add extra requires for odo

llllllllll:
https://github.com/blaze/blaze/issues/1600


.travis.yml
```diff
@@ -41,21 +41,18 @@ install:
...
-  - conda create -n odo python=$TRAVIS_PYTHON_VERSION pytest numpy pandas sqlalchemy
-  - conda install -n odo h5py pip cython bcolz coverage networkx toolz multipledispatch
-  - conda install -n odo pytables paramiko boto boto3 bokeh pymysql
-  - if [ -n "$PANDAS_VERSION" ]; then conda remove -n odo pandas; fi
-  - source activate odo
+  # Use conda **ONLY** for numpy and pandas, this speeds up the builds a
+  # lot. Use the normal pip install for the rest.
+  - conda create -n odo python=$TRAVIS_PYTHON_VERSION pandas=${PANDAS_VERSON:-0.19.0} numpy=1.11.2
+  # update setuptools and pip
+  - pip install -u setuptools pip
```
richafrank: Does `-u` work for you? Only `-U` works for me...

richafrank: Looks like travis agrees with me 😄

setup.py
```diff
@@ -26,6 +26,37 @@ def read(filename):
         return f.read()
 
 
+def read_reqs(filename):
+    return read(filename).strip().splitlines()
+
+
+install_requires = read_reqs('etc/requirements.txt')
```
richafrank: No concerns about the behavior, but do you think it would be clearer for people if `install_requires` and `extras_require` were both functions or both lists of requirements?

setup.py
```diff
@@ -26,6 +26,37 @@ def read(filename):
...
+    postgres = sql + read_reqs('etc/requirements_postgres.txt')
+    pytables = read_reqs('etc/requirements_pytables.txt')
+    sas = read_reqs('etc/requirements_sas.txt')
+    ssh = read_reqs('etc/requirements_ssh.txt')
+    test = read_reqs('etc/requirements_test.txt')
+
+    extras = locals()
+    extras['all'] = list(chain.from_iterable(extras.values()))
+    # write the ci target after collecting all of the others so that the ci
+    # frozen dependencies don't trample the less strict versions.
```
richafrank: Instead of relying on no unexpected locals being introduced, what do you think of:
```python
def extras_require():
    extras = {
        req: read_reqs('etc/requirements_{req}'.format(req=req)) 
        for req in {
          'sql', 'aws', 'bcolz', 'bokeh', 'dask', 'h5py', 'mongo', 'mysql', 'postgres',
          'pytables', 'sas', 'ssh', 'test', 'ci',
        }
    }
    extras['mysql'] += extras['sql']
    extras['postgresql'] += extras['sql']
    return extras
```

setup.py
```diff
@@ -36,8 +67,8 @@ def read(filename):
       license='BSD',
       keywords='odo data conversion hdf5 sql blaze',
       packages=find_packages(),
-      install_requires=read('requirements.txt').strip().split('\n'),
-      extra_requires={'all': read('extra-requirements.txt').strip().split('\n')},
+      install_requires=install_requires,
+      extras_require=extras_require(),
```
richafrank: I've seen this mistake (that you fixed here) many times before...

.travis.yml
```diff
@@ -41,21 +41,19 @@ install:
   - conda update conda
 
   # Install dependencies
-  - conda create -n odo python=$TRAVIS_PYTHON_VERSION pytest numpy pandas sqlalchemy
-  - conda install -n odo h5py pip cython bcolz coverage networkx toolz multipledispatch
-  - conda install -n odo pytables paramiko boto boto3 bokeh pymysql
-  - if [ -n "$PANDAS_VERSION" ]; then conda remove -n odo pandas; fi
+  # Use conda **ONLY** for numpy and pandas, this speeds up the builds a
+  # lot. Use the normal pip install for the rest.
+  - conda create -n odo python=$TRAVIS_PYTHON_VERSION pandas=${PANDAS_VERSON:-0.19.0} numpy=1.11.2
```
richafrank: When is the pandas default version used?

richafrank: When the time comes to update the numpy version for ci, someone will have to do it twice (once here and once in reqs). Do you want to remove it from reqs, if it's really getting installed here?

.travis.yml
```diff
@@ -41,21 +41,19 @@ install:
...
   - source activate odo
+  # update setuptools and pip
+  - conda update setuptools pip
+
+  # install the frozen ci dependencies
+  - pip install -e .[ci]
```
richafrank: Nice and clean!

.travis.yml
llllllllll: I like that users can just pip install and go on their local systems, what do you think about putting a comment in the reqs file next to pandas and numpy to tell users to update the .travis file?

setup.py
llllllllll: yeah, I think I need to do that in blaze also so I will make this a func.

setup.py
llllllllll: that's clean, thanks!

setup.py
llllllllll: yeah, I never remember which is plural.

.travis.yml
richafrank: > I like that users can just pip install and go on their local systems, what do you think about putting a comment in the reqs file next to pandas and numpy to tell users to update the .travis file?

:+1:

richafrank: > The default pandas version is used when we are not pulling from github

What do you think about putting the default in the travis matrix too, so whether it's coming from github or pypi, that information is all in the same place?

llllllllll: sure

PR #499, opened 2016-11-10
Fix test regressions

kwmsmith:


necaris:
@kwmsmith thank you for this! I don't see any changes in the diff?

PR #501, opened 2016-11-15
BUG: fix postgres to/from dataframe quoting

llllllllll:


odo/backends/sql.py
```diff
@@ -692,6 +692,7 @@ def select_or_selectable_to_frame(el, bind=None, dshape=None, **kwargs):
             parse_dates=datetime_fields,
             dtype=other_dtypes,
             skip_blank_lines=False,
+            escapechar=kwargs.get('escapechar', '\\'),
```
richafrank: Had to find [the docs] (http://pandas.pydata.org/pandas-docs/version/0.18.1/generated/pandas.read_csv.html) to understand this better, but now...
```
escapechar : str (length 1), default None

    One-character string used to escape delimiter when quoting is QUOTE_NONE.
```
```
quoting : int or csv.QUOTE_* instance, default 0

    Control field quoting behavior per csv.QUOTE_* constants. Use one of QUOTE_MINIMAL (0), QUOTE_ALL (1), QUOTE_NONNUMERIC (2) or QUOTE_NONE (3).
```

Does this mean that the `escapechar` isn't used unless `QUOTE_NONE`, i.e. not the default and therefore not what's used here, is used?

odo/backends/sql_csv.py
```diff
@@ -322,14 +324,30 @@ def append_dataframe_to_sql_table(tbl,
             df[[c.name for c in tbl.columns]],
             buf,
             index=False,
+            quotechar=quotechar,
+            doublequote=True,
+            # use quotechar for escape intentionally because it makes it makes
```
richafrank: typo - repeated "it makes"

odo/backends/sql.py
llllllllll: This may not be needed anymore. I was playing around with `QUOTE_NONNUMERIC` but that breaks NaT timestamps. I will double check.

llllllllll: This is used to handle quoting of the quotechar even when quoting is set to `QUOTE_MINIMAL`. We tell postgres to use this same quotechar so it escapes the quotes with a backslash (by default) when generating the csv. If this is removed test to and from dataframe string fail.

PR #505, opened 2016-11-17
Add dask as hard requirement

necaris:
With the work on the `chunks` backend merged in #481 , `dask` becomes a hard requirement.

@llllllllll would appreciate a sanity check on this ASAP so I can tag `0.5.2` and build packages; this is a prerequisite for a similar tag-and-update need on Blaze.

### DIFF
--- a/etc/requirements.txt
+++ b/etc/requirements.txt
@@ -4,3 +4,4 @@ pandas >= 0.15.0
 toolz >= 0.7.3
 multipledispatch >= 0.4.7
 networkx >= 1.0
+dask >= 0.11.1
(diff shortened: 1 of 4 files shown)


llllllllll:
Should we remove the dask extras install target in setup.py since it will always be installed? idk if there are other requirements in requirements_dask.py

necaris:
I agree, will add to this PR

--
Rami Chowdhury
"A mind all logic is like a knife all blade -- it makes the hand bleed that
uses it." -- Rabindranath Tagore

১৭ নভেম্বর, ২০১৬ ৬:৫৪ pm এ, "Joe Jevnik" <notifications@github.com> লিখেছেন:

etc/requirements.txt
```diff
@@ -4,3 +4,4 @@ pandas >= 0.15.0
 toolz >= 0.7.3
 multipledispatch >= 0.4.7
 networkx >= 1.0
+dask >= 0.11.1
```
llllllllll: should this still be `dask[complete]` to get the reqs for dask dataframe and stuff?

PR #507, opened 2016-11-23
VCF format

Horta:
Support for widely used file format for genotypes.

http://www.internationalgenome.org/wiki/Analysis/Variant%20Call%20Format/vcf-variant-call-format-version-40/

### DIFF
--- a/odo/backends/tests/test_vcf.py
+++ b/odo/backends/tests/test_vcf.py
@@ -1,26 +1,20 @@
 from __future__ import absolute_import, division, print_function
+from __future__ import unicode_literals
 
 import pytest
 
 import sys
 import os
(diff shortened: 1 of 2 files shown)


llllllllll:
Thank you for taking the time to write this! It looks like a `VCF` file is just a special kind of CSV. I think it may be easier to implement this `VCF` as a subclass of `CSV` which forwards the seperator and comment chars to the super class. Then you would need to pass around the `VCF` object instead of just the file URI but it would simplify the implementation a lot. I am also not sure if this format is general enough to be put into odo directly because it is just a special case of CSV. One option would be to create a third-party package which implements your type and adds it to the odo graph so that users may download your extension if needed.

PR #508, opened 2016-11-24
Tidying up

Horta:
I'm exporting now only the names that are actually used by the end-user, which makes easier
to getting started with odo (think about ipython and tab-completion).

Probably didn't get everything right now but I will revisit this again.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,0 @@
-odo/_version.py export-subst
--- a/odo/backends/ssh.py
+++ b/odo/backends/ssh.py
@@ -57,7 +57,8 @@ class _SSH(object):
     Examples
     --------
 
-    >>> from odo import SSH, CSV
+    >>> from odo.backends.ssh import SSH
+    >>> from odo.backends.csv import CSV
     >>> s = SSH(CSV)('/path/to/file.csv', hostname='hostname', username='alice')
 
     Normally create through resource uris


odo/__init__.py
```diff
@@ -1,73 +1,37 @@
-from __future__ import absolute_import, division, print_function
+from __future__ import absolute_import as _
+from __future__ import division as _
+from __future__ import print_function as _
+
+from pkg_resources import DistributionNotFound as _DistributionNotFound
+from pkg_resources import get_distribution as _get_distribution
+
+try:
+    __version__ = _get_distribution('odo').version
```
llllllllll: we use versioneer to manage this variable, this should be restored to what it was before.

odo/backends/__init__.py
```diff
@@ -0,0 +1,10 @@
+from importlib import import_module as _imod
```
llllllllll: I read this as operator.imod, can you just call this `_import_module`?

odo/__init__.py
```diff
@@ -1,73 +1,37 @@
...
-with ignoring(ImportError):
-    from .backends.hdfstore import HDFStore
-with ignoring(ImportError):
-    from .backends.pytables import PyTables
-with ignoring(ImportError):
-    from .backends import sql
-with ignoring(ImportError):
-    from .backends import mongo
-with ignoring(ImportError):
-    from .backends.csv import CSV
```
llllllllll: I think all of the types we define should still be part of the top level API, thinks like `odo.CSV` are very useful and we shouldn't break code that currently depends on those names being there.

odo/backends/dask.py
```diff
@@ -12,12 +12,14 @@
...
-from odo import append, chunks, convert, discover, TextFile
+from odo import append, discover, convert
+from odo.chunks import chunks
+from odo.backends.text import TextFile
 from ..utils import filter_kwargs
 
 
 @discover.register(Array)
-def discover_dask_array(a, **kwargs):
+def discover_dask_array(a, **_):
```
llllllllll: this change is not needed, this pattern is used all over and it is expected that `**kwargs` may or may not be used.

setup.py
```diff
@@ -1,24 +1,22 @@
-#!/usr/bin/env python
 from itertools import chain
```
llllllllll: can you explain the reason for the changes in this file? I don't think people should run the tests through the setup.py file so we can probably remove the test requires stuff.

.travis.yml
```diff
@@ -74,7 +74,7 @@ before_script:
 
 script:
   - conda list
-  - py.test -v --doctest-modules --doctest-ignore-import-errors -rs odo
```
llllllllll: This is not how most people run their test suite so I would prefer to keep it as it is.

odo/__init__.py
```diff
@@ -1,73 +1,37 @@
-from __future__ import absolute_import, division, print_function
+from __future__ import absolute_import as _
```
llllllllll: I don't think this is needed. Many projects export these names without confusing people, for example: numpy does this.

odo/directory.py
```diff
@@ -26,6 +26,7 @@ class _Directory(Chunks):
 
     """
     def __init__(self, path, **kwargs):
+        super(_Directory, self).__init__(None)
```
llllllllll: why are we forwarding  `None` to the constructor now?

odo/directory.py
```diff
@@ -46,11 +47,11 @@ def Directory(cls):
     re_path_sep = '\\\\'
 
 @discover.register(_Directory)
-def discover_Directory(c, **kwargs):
+def discover_Directory(c, **_):
```
llllllllll: this can stay `**kwargs`

odo/directory.py
```diff
@@ -46,11 +47,11 @@ def Directory(cls):
...
     return var * discover(first(c)).subshape[0]
 
 
-@resource.register('.+' + re_path_sep + '\*\..+', priority=15)
+@resource.register(r'.+' + re_path_sep + r'\*\..+', priority=15)
```
llllllllll: +1, thanks

Issue #151, opened 2015-03-26
Directory of gzipped CSVs is broken


cpcloud:
```
mkdir test
cd test
$ echo -en "a,b\n1,2\n3,4" > a.csv
$ echo -en "a,b\n1,2\n3,4\n5,6" > b.csv
$ gzip < *.csv > ab.csv.gz
$ cd ..
$ ipython
In [1]: df = odo('test/*.csv.gz', pd.DataFrame)
KeyError: <class 'odo.directory.Directory(CSV)'>
```

cpcloud:
Hm, even though `Directory` inherits from `Chunks` since `convert` doesn't respect inheritance, we need to explicitly define this conversion.

ethanluoyc:
This is still occurring on version 0.5.0 (the latest development version?)
It can be reproduced with the steps mentioned in #153. 

I would like to help fix this one. Where are we right now on this and what can I do?

PR #516, opened 2016-12-16
Fix convert a directory of files

ethanluoyc:
This PR Addresses #151.

Basically, the bug happens as the the dynamically generated parameterized Directory class is not found in the graph in NetworkDispatcher. A conversion path should be identified by the parameterized Chunks superclass.

I still need to add test cases so that this bug will not reappear, but where is best place for me to add the test cases?

### DIFF
--- a/odo/core.py
+++ b/odo/core.py
@@ -158,6 +158,12 @@ def path(graph, source, target, excluded_edges=None, ooc_types=ooc_types):
                 source = cls
                 break
 
+    # For Directory class we need its container type to find the path in the graph
+    from odo.chunks import chunks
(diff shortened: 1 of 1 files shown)


llllllllll:
Thank you for working on this. This looks like a pretty annoying issue with subclasses in general so I decided to generalize `NetworkDispatcher` (`convert`) to search the entire mro of the source type. Then, the `Directory` fix falls out of that: #518. Do you want to look over that change to make sure it solves your problem?

ethanluoyc:
I think that's a better approach. I will close this PR in favor of #518.

Issue #515, opened 2016-12-15
Add variants of SQL types


postelrich:
Came across this issue where I was trying to use blaze on a sqlalchemy table I created with a column made with the `sa.Numeric` type.  It yielded an error, and when I investigated saw that blaze looks for `sa.NUMERIC` instead.  Not quite sure if there is a difference, but if not we should add the lower case to the type checking.

```
t = sa.Table(name, meta,  sa.Column('foo', sa.Numeric(precision=18), primary_key=True))
bz.data(t)
```

postelrich:
line in question: https://github.com/blaze/odo/blob/master/odo/backends/sql.py#L187

llllllllll:
I wonder if we want to swap to _just_ `Numeric`:

```
In [2]: sa.Numeric.mro()
Out[2]: 
[sqlalchemy.sql.sqltypes.Numeric,
 sqlalchemy.sql.sqltypes._DateAffinity,
 sqlalchemy.sql.type_api.TypeEngine,
 sqlalchemy.sql.visitors.Visitable,
 object]

In [3]: sa.NUMERIC.mro()
Out[3]: 
[sqlalchemy.sql.sqltypes.NUMERIC,
 sqlalchemy.sql.sqltypes.Numeric,
 sqlalchemy.sql.sqltypes._DateAffinity,
 sqlalchemy.sql.type_api.TypeEngine,
 sqlalchemy.sql.visitors.Visitable,
 object]
```

the lowercase type will catch both, but like you I am not sure what the difference is.

necaris:
👍 for just swapping to `Numeric` here since `NUMERIC` doesn't seem to buy us anything?

llllllllll:
Is this a problem with other types, we may as well fix this in one pass.

PR #519, opened 2016-12-19
Fix for #515 switch to Numeric type

postelrich:
There are some other types that could be changed though it would mean changing the way we do the lookup.  Plus some things like `FLOAT` are children of `Numeric`.  Not sure how we want to procede here. 

```
In [4]: sa.DECIMAL.mro()
Out[4]:
[sqlalchemy.sql.sqltypes.DECIMAL,
 sqlalchemy.sql.sqltypes.Numeric,
 sqlalchemy.sql.sqltypes._DateAffinity,
 sqlalchemy.sql.type_api.TypeEngine,
 sqlalchemy.sql.visitors.Visitable,
 object]

In [5]: isinstance(sa.DECIMAL(), sa.Numeric)
Out[5]: True

In [6]: sa.DATETIME.mro()
Out[6]:
[sqlalchemy.sql.sqltypes.DATETIME,
 sqlalchemy.sql.sqltypes.DateTime,
 sqlalchemy.sql.sqltypes._DateAffinity,
 sqlalchemy.sql.type_api.TypeEngine,
 sqlalchemy.sql.visitors.Visitable,
 object]

In [7]: sa.TIMESTAMP.mro()
Out[7]:
[sqlalchemy.sql.sqltypes.TIMESTAMP,
 sqlalchemy.sql.sqltypes.DateTime,
 sqlalchemy.sql.sqltypes._DateAffinity,
 sqlalchemy.sql.type_api.TypeEngine,
 sqlalchemy.sql.visitors.Visitable,
 object]

In [2]: sa.Float.mro()
Out[2]:
[sqlalchemy.sql.sqltypes.Float,
 sqlalchemy.sql.sqltypes.Numeric,
 sqlalchemy.sql.sqltypes._DateAffinity,
 sqlalchemy.sql.type_api.TypeEngine,
 sqlalchemy.sql.visitors.Visitable,
 object]
```

Maybe we just want to add more types to `revtypes` so we support more types and maybe a better explanation in the NotImplementedError so they know where to look for supported types.

### DIFF
--- a/odo/backends/sql.py
+++ b/odo/backends/sql.py
@@ -184,7 +184,7 @@ def discover_typeengine(typ):
         return dshape(revtypes[typ])[0]
     if type(typ) in revtypes:
         return revtypes[type(typ)]
-    if isinstance(typ, (sa.NUMERIC, sa.DECIMAL)):
+    if isinstance(typ, sa.Numeric):
         return datashape.Decimal(precision=typ.precision, scale=typ.scale)
     if isinstance(typ, (sa.String, sa.Unicode)):
         return datashape.String(typ.length, 'U8')


llllllllll:
Should we just add more `discover` dispatches instead of having a big if/else statement in the dispatch of `TypeEngine`? Basically refactor the current branches (with your change) into new `dispatch(type)(discover)` functions so that we can leverage the good mro that sqlalchemy has.

postelrich:
I think we're good with just this.  I see that `revtypes` does have `DateTime` and `Float`.  I think this was just an outlier case where the base Numeric type wasn't being handled.

llllllllll:
Okay, do you want to add a regression test to `test_sql.py` so that if I decide to do that refactor later we know this still works?

PR #521, opened 2016-12-28
BUG: fix problematic string in postgres->dataframe

llllllllll:
the string `nan` would be returned as a float object. This preserves all string values. This also ensures quote characters and escapes are passed through correctly.

### DIFF
--- a/odo/backends/tests/test_postgres.py
+++ b/odo/backends/tests/test_postgres.py
@@ -383,17 +383,44 @@ def test_to_dataframe_strings(sql_with_strings):
     sql_with_strings, bind = sql_with_strings
 
     insert_query = sql_with_strings.insert().values([
-        {'non_optional': 'ayy', 'optional': 'hello "world"'},
-        {'non_optional': 'lmao', 'optional': None},
+        # quoting is done correctly
+        {'non_optional': 'hello "world"', 'optional': 'hello "world"'},
+
+        # None stays None
+        {'non_optional': 'ayy lmao', 'optional': None},
+
+        # nan string is just the string nan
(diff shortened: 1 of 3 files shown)


necaris:
👍

llllllllll:
I am seeing test failures in an application that uses this. I will see if it is due to workarounds for this bug or a real issue and add tests where needed.

odo/backends/tests/test_postgres.py
```diff
@@ -383,17 +383,40 @@ def test_to_dataframe_strings(sql_with_strings):
...
     ])
     if bind is None:
         insert_query.execute()
     else:
         bind.execute(insert_query)
 
+    expected = pd.DataFrame(
+        [['ayy', 'hello "world"'],
+         ['lmao', None],
+         ['nan', 'this ticker sucks'],
```
ehebert: We discussed in person about testing `nan` and `NULL` on both sides of the test, other than that, LGTM.
If test failures you are seeing on the dependent project require more work, let me know and can review those changes.

llllllllll:
This is fixed now, we use a sentinel na_value on the db and csv reader. I have a more long term solution that should be faster and more robust but this is good for now.

odo/backends/tests/test_postgres.py
```diff
@@ -383,17 +385,88 @@ def test_to_dataframe_strings(sql_with_strings):
     sql_with_strings, bind = sql_with_strings
 
     insert_query = sql_with_strings.insert().values([
```
mtydykov: Should there be a case for each of the items in `_default_na`?

PR #525, opened 2017-01-03
ENH: MSSQL tests/support

thequackdaddy:
Hello,

So I've made some progress on getting MSSQL work with this. This is nowhere near ready. Got a couple of preference-type questions:

* There are (at least) [5 different ways](http://docs.sqlalchemy.org/en/latest/dialects/mssql.html#module-sqlalchemy.dialects.mssql.pyodbc) to connect to MSSQL from sqlalchmey. I've really only played with pyodbc--I've had generally good luck with ODBC in general on windows and it works in python. So I plan to focus on the `mssql+pyodbc` connection.  UPDATE: This now uses `mssql+pymssql` because getting username, password, and trusted connection is more obvious to me.
* To bulk import/export data there are really 3 choices ([here's a more detailed breakdown](https://msdn.microsoft.com/en-us/library/ms175937.aspx))
  * The command line `bcp` utility. It is installed with SQL Server. Downsides to the `bcp` utility is that it is entirely run from the command prompt and you have to provide a server name. If you are using ODBC (or an ODBC DSN), I'm unsure how to get the servername from the ODBC connection. 
  * The `BULK INSERT` command which can be run from the TSQL interface. The downsides are that you must provide a file path which is a server-side read. So if the *.csv file is on a client computer, getting the SQL server to see the file and upload it can be a challenge. Also, the `BULK INSERT` is an "insert-only" command. There's no `BULK EXPORT` to csv functionality (which odo uses is the `sql.py` file, so we have to use a fallback through the iterators. 
  * The `OPENROWSET` statement in an `INSERT`. Basically, the same benefits (and drawbacks) of `BULK INSERT`. 
* I'm using bcp.exe and deriving the parameters from `pymssql`. If there was a way to get the connection parameters (servername, username, password, etc.) from `pyodbc`, that would be great. Open for ideas. 

ToDo:

* [x] bit support: not that i had to edit the `unit_to_dtype` code so that `?bool` -> `'bool`. This appears to coerce `NULL` bools to `False`.
* [x] money support
* [x] bcp export doesn't do headers. I think creating 2 files and `copy /b` them together is the way to go. 
* [x] bcp result message parsing (currently ignoring everything) - Error parsing is a little rough. SQL server likes to give warnings that blanks become nulls...
* [x] not sure if pipe (|) delimiters work (I use them a lot... I tend to have commas in data, e.g., street address of 123 Main St., Apt 3)
* [ ] need a reviewer (I'll confess I'm not a professional programmer)

MaybeDo:
* [ ] datetimeoffset support (? this is kind of hard, I think)
* [ ] pyodbc support through connection string
* [ ] add pymssql or pyodbc to requirements/setup.py
* [ ] fallback to bulk insert, then standard `INSERT` if bcp insert won't work (maybe if pyodbc used and it doesn't make sense)
* [ ] fallback to sqlalchemy export if bcp won't work. 
* [ ] For BCP, I'm a little unclear which code page or encoding thing I should use. Seems like `-c` seems like the best. I'm confused about the others.

Thanks!

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,7 +15,6 @@ platform:
     - x64
 
 services:
-  # - mssql2014 # No tests for mssql written yet
   # - mysql # mysql tests fail. Don't understand how to fix at the moment.
   - postgresql93
   - mongodb
@@ -51,6 +50,14 @@ build_script:
 
   # install python stuff
   - sh etc\ci-install.sh
+
+  # MSSQL
+  # pymssql requires TCP/IP
+  - SET INSTANCENAME=SQL2014
+  - net start MSSQL$%INSTANCENAME%
+  - "powershell ci\\sql_server_tcpip.ps1"
+  - sqlcmd -E -Q "CREATE DATABASE test;"
+
 
 test_script:
   - activate odo
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -50,6 +50,8 @@ build_script:
 
   # install python stuff
   - sh etc\ci-install.sh
+  - conda install cython
+  - pip install git+https://github.com/pymssql/pymssql/
 
   # MSSQL
   # pymssql requires TCP/IP
@@ -58,7 +60,5 @@ build_script:
   - "powershell ci\\sql_server_tcpip.ps1"
   - sqlcmd -E -Q "CREATE DATABASE test;"
 
-
 test_script:
-  - activate odo
   - py.test -v --doctest-modules --doctest-ignore-import-errors -rs odo
--- a/ci/sql_server_tcpip.ps1
+++ b/ci/sql_server_tcpip.ps1
@@ -1,0 +1,20 @@
+[reflection.assembly]::LoadWithPartialName("Microsoft.SqlServer.Smo") | Out-Null
+[reflection.assembly]::LoadWithPartialName("Microsoft.SqlServer.SqlWmiManagement") | Out-Null
+
+$serverName = $env:COMPUTERNAME
+$instanceName = $env:INSTANCENAME
+$smo = 'Microsoft.SqlServer.Management.Smo.'
+$wmi = new-object ($smo + 'Wmi.ManagedComputer')
+
+# Enable TCP/IP
+$uri = "ManagedComputer[@Name='$serverName']/ServerInstance[@Name='$instanceName']/ServerProtocol[@Name='Tcp']"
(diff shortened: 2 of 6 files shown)


thequackdaddy:
OK all tests are green except an unrelated dask. A few updates...

* After struggling with ODBC and BULK INSERT, I switched to essentially writing a wrapper for bcp.exe. 
* I am using pymssql instead of pyodbc. pyodbc will still "kind of" work, but the direct interface to bcp.exe requires pymssql
* Still have to check that `bit`, `money`, and a few other mssql-specific dtypes work. 

I appreciate any commentary about this.

llllllllll:
ack, I will try to look over this sometime this weekend.

thequackdaddy:
I had forgotten how fast loading data with `bcp.exe` can be:

```python
In [2]: %%time
   ...: 
   ...: from odo import odo 
   ...: import blaze as bz
   ...: nyc = odo('yellow_tripdata_2016-01.csv',
   ...:           'mssql+pymssql://localhost/test::yellow201601')
   ...: 
   ...: 

Wall time: 2min 31s


In [3]: bznyc = bz.data(nyc)
   ...: bznyc.count()
Out[3]: 10906858
```

thequackdaddy:
ping to see if anyone has time to review/comment over the next few weeks

thequackdaddy:
Pinging again to see if anyone has interest in picking this up. 

I think there the CI failure is an unrelated (says something about dask)

Thanks.

FedericoFontana:
Any news? It would be great to have MSSQL support.

thequackdaddy:
I guess I'm still waiting for input from others. 

The one (kind of) big challenge is how bulk inserts should work. I think using `bcp.exe` makes the most sense, but you have to provide it with login arguments. You would expect something like `mssql+pyodbc://user:password@server::table`, but that doesn't nicely handle when you want to use a trusted connection. (Which is all I use). 

Additionally, #568 is a new quirk. MSSQL's `timestamp` type is different than other platforms. A bug was introduced to handle this, but it turns out the fix was too simple and too general and causes a testcase fail in blaze itself. 

Happy to help get this working, but I'm not well-versed enough in how this *should* work to do it alone. 

In the mean time, Pulling this PR works well enough for my purposes.

rvaidya:
+1 SQL Server support would be great, am currently running into this and will most likely need to use pandas directly.

PR #532, opened 2017-01-08
Fix syntax for sqlalchemy-redshift CopyCommand

mharradon:
I'm not sure why this line uses the keyword argument "table" - I can't find any docs, past or present, of CopyCommand using that syntax. This fixes an error I had when copying csv to redshift.

### DIFF
--- a/odo/backends/sql_csv.py
+++ b/odo/backends/sql_csv.py
@@ -197,7 +197,7 @@ else:
 
         compression = getattr(element, 'compression', '').upper() or None
         cmd = CopyCommand(
-            table=element.element,
+            to=element.element,
             data_location=element.csv.path,
             access_key_id=aws_access_key_id,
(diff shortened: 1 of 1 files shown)


necaris:
Note that it seems [version 0.3 and earlier _did_ use `table`](https://github.com/sqlalchemy-redshift/sqlalchemy-redshift/blob/bd9b13fed3165ce8d4d4f1a710c746ace3053145/redshift_sqlalchemy/commands.py#L216). I don't see any reason why we should be tied to this old version, but we should either handle both argument names or explicitly [require a particular version](https://github.com/blaze/odo/blob/master/extra-requirements.txt#L9).

mharradon:
If we just use a positional argument instead of a keyword argument it should be backwards compatible.

Alternately if odo requires a sqlalchemy-redshift version that's 2 releases behind it should probably be in the requirements.

PR #533, opened 2017-01-08
WIP/ENH: bcolz coerce to ascii

thequackdaddy:
Hello,

So I have another idea... this one I can imagine might not fly. 

Long story short, is I have data in a MSSQL server that I'm trying to get into bcolz. The MSSQL PR #525 will let me pull from the server pretty nicely. 

The next problem I've run into is that the data in MSSQL is essentially unicode, but bcolz (and numexpr) don't like unicode--they prefer strings. So much so that by default when you run the `bcolz.fromdataframe` command, it attempts to coerce  all strings to  ascii/bytes as well. 

I propose to  make that conversion a little easier... Here's an idea. 

Let me know if this is useful. 

An example of why this might be useful... 

I have a field of vehicle types in a SQL table... think `F-150`, `COROLLA`, `CAMRY`,  etc. When I use `odo` to move this to bcolz,, 100,000 records becomes 1235 separate *. blp files. Using this, I can cut that down to 306 files.

### DIFF
--- a/odo/backends/bcolz.py
+++ b/odo/backends/bcolz.py
@@ -23,14 +23,14 @@ def discover_bcolz(c, **kwargs):
 
 
 @append.register((ctable, carray), np.ndarray)
-def numpy_append_to_bcolz(a, b, **kwargs):
+def numpy_append_ndarray_to_bcolz(a, b, **kwargs):
     a.append(b)
     a.flush()
     return a
 
 
 @append.register((ctable, carray), object)
-def numpy_append_to_bcolz(a, b, **kwargs):
+def numpy_append_object_to_bcolz(a, b, **kwargs):
     return append(a, convert(chunks(np.ndarray), b, **kwargs), **kwargs)
 
 
@@ -81,7 +81,18 @@ def resource_bcolz(uri, dshape=None, expected_dshape=None, **kwargs):
                              " valid datashape")
         dshape = datashape.dshape(dshape)
 
+        # bcolz don't like unicode (yet). Should we coerce unicode strings
+        # to ascii (bytes)?
+        coerce = kwargs.get('bcolz_coerce_unicode', False)
+        if coerce:
+            if datashape.predicates.isrecord(dshape.measure):
+                for i, ty in enumerate(dshape.parameters[1].types):
+                    ty = getattr(ty, 'ty', ty)
+                    if hasattr(ty, 'encoding'):
+                        ty.encoding = 'A'
+
         dt = dshape_to_numpy(dshape)
+
         shape_tail = tuple(map(int, dshape.shape[1:]))  # tail of shape
         if dshape.shape[0] == datashape.var:
             shape = (0,) + shape_tail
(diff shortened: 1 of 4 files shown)


odo/backends/bcolz.py
```diff
@@ -81,7 +81,18 @@ def resource_bcolz(uri, dshape=None, expected_dshape=None, **kwargs):
...
 
+        # bcolz don't like unicode (yet). Should we coerce unicode strings
+        # to ascii (bytes)?
+        coerce = kwargs.get('bcolz_coerce_unicode', False)
+        if coerce:
+            if datashape.predicates.isrecord(dshape.measure):
+                for i, ty in enumerate(dshape.parameters[1].types):
+                    ty = getattr(ty, 'ty', ty)
+                    if hasattr(ty, 'encoding'):
+                        ty.encoding = 'A'
```
llllllllll: I like the idea here buy I don't think we should be mutating the dshape in place. Can we do a functional update and forward the newly encoded dshape along?

odo/backends/bcolz.py
```diff
@@ -81,7 +81,18 @@ def resource_bcolz(uri, dshape=None, expected_dshape=None, **kwargs):
                              " valid datashape")
         dshape = datashape.dshape(dshape)
 
+        # bcolz don't like unicode (yet). Should we coerce unicode strings
+        # to ascii (bytes)?
+        coerce = kwargs.get('bcolz_coerce_unicode', False)
+        if coerce:
+            if datashape.predicates.isrecord(dshape.measure):
+                for i, ty in enumerate(dshape.parameters[1].types):
```
llllllllll: I think `i` is unused

llllllllll: I think we can just use`dshape.fields`

odo/backends/tests/test_bcolz.py
```diff
@@ -189,3 +189,33 @@ def test_csv_to_bcolz():
         with tmpfile('bcolz') as tgt:
             bc = into(tgt, src)
             assert len(bc) == 3
+
+
+def test_coerce_ascii():
+    # The point of this test is to check that the bcolz_coerce_unicode
+    # option works
+    import datashape
```
llllllllll: ket's move these to the top of the file

odo/convert.py
```diff
@@ -195,7 +196,20 @@ def list_to_numpy(seq, dshape=None, **kwargs):
     if (seq and isinstance(seq[0], Iterable) and not ishashable(seq[0]) and
             not isscalar(dshape)):
         seq = list(map(tuple, seq))
-    return np.array(seq, dtype=dshape_to_numpy(dshape))
+    try:
+        return np.array(seq, dtype=dshape_to_numpy(dshape))
+    except UnicodeEncodeError:
+        ds = dshape_to_numpy(dshape)
+        for i, rec in enumerate(seq):
+            if sys.version_info.major == 3:
```
llllllllll: it looks like the only difference in these branches is the `str` vs `unicode` call. We could just assign a `string_type` function in this branch and then share the expression in both cases.

thequackdaddy:
Hello. 

I went back to playing with this, and I'm thinking this kind of feature would be more trouble than its worth. 

I'll close.

PR #546, opened 2017-05-25
pd.NaTType is removed in pandas 0.20

timcera:
In pandas 0.20 there is no longer a "pd.NaTType".  Replaced with "type(pd.NaT)"


llllllllll:
the build failure looks unrelated, I will try to fix that but the change looks good to me

jasonmellone:
It seems pypi is behind the repo. This should be fixed.

PR #553, opened 2017-07-07
Fix usage of deprecated pandas api

dhirschfeld:
The `tslib` namespace is [`deprecated`](https://github.com/pandas-dev/pandas/pull/16146).

Calling `type(pd.NaT)` is equivalent and works even on pandas `0.15.0`.

### DIFF
--- a/odo/backends/pandas.py
+++ b/odo/backends/pandas.py
@@ -99,7 +99,7 @@ def nan_to_nat(fl, **kwargs):
     raise NotImplementedError()
 
 
-@convert.register((pd.Timestamp, pd.Timedelta), (pd.tslib.NaTType, type(None)))
+@convert.register((pd.Timestamp, pd.Timedelta), (type(pd.NaT), type(None)))
 def convert_null_or_nat_to_nat(n, **kwargs):
     return pd.NaT
 


llllllllll:
idk why this test is flapping, it looks like it is trying to hit some remote resource. We need to patch that so its not actually doing network IO. I probably won't be able to do that until this weekend but I could help you with it if you want to fix the test ;)

dhirschfeld:
Yeah, it seems a of an anti-pattern having unit-tests trying to hit the web.

If you don't get to it first I can take a look next week as it's coming up :beer: o'clock down these parts!

dhirschfeld:
The offending test is:
```python
@pytest.mark.xfail(raises=HTTPError)
def test_failed_url():
    failed_url = "http://foo.com/myfile.csv"
    with tmpfile('.csv') as fn:
        odo(failed_url, fn)
```
When I run that locally it works fine but on Travis it seems to give a timeout error rather than an HTTPError:
```
URLError: <urlopen error [Errno 110] Connection timed out>
```

dhirschfeld:
I don't have much experience mocking http requests but I gave it a go over at #557.

dhirschfeld:
See also #546

dhirschfeld:
@llllllllll - would be good to get this merged if you can find the time...

llllllllll:
is this the same as: https://github.com/blaze/odo/pull/546/files?

dhirschfeld:
Duplicate of #546

Issue #571, opened 2017-09-01
Bulk Loaders and Exports


AlexHagerman:
Hi,

I've used odo in some of my smaller projects and enjoy it, but with others the current data migrations I do use pyodbc and calls to utilities like [bcp](https://docs.microsoft.com/en-us/sql/tools/bcp-utility) and [nzload](https://www.ibm.com/support/knowledgecenter/SSULQD_7.1.0/com.ibm.nz.load.doc/c_load_nzload_command.html) due to the size of the data I am importing/exporting.

I don't see much documentation around utilities like these in Odo, but I did see that [pg_bulkload](http://odo.pydata.org/en/latest/perf.html) was not implemented yet.

I was wondering if there was any movement in this space and if there is on boarding documentation for somebody that might like to assist or work on some contributions relating to these bulk import/export utilities? If that doesn't fit the project any ideas or suggestions for somebody interested in creating an extension library for Odo with these utilities?

Thanks.

dhirschfeld:
I've previously used `bcp` by writing csvs to spinning rust but whilst it does get you very good speedups the whole process felt a bit icky.  In contrast the postgres bulk upload could be used entirely in memory by writing the csv to a `StringIO`. I tried to avoid writing to disk by using a named pipe but couldn't get that working. I've read various things on the internet that suggest there *may* be a way to make that work but I haven't followed that up.

In this instance I've got a bit of `.NET` envy - the [`SqlBulkCopy`](https://docs.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlbulkcopy?view=netframework-4.7) class uses the same apis as `bcp` and so gets the same performance but directly from C# code. I haven't pursued the named pipe route as I think the proper way to do this is to implement the [Tabular Data Stream Protocol](https://msdn.microsoft.com/en-us/library/cc448435.aspx?f=255&MSPPError=-2147217396) directly in Python as `SqlBulkCopy` does in C#. I think that doable but is a big project which I don't personally have time for at the moment.

***Refs:***
* https://msdn.microsoft.com/en-us/library/cc448435.aspx?f=255&MSPPError=-2147217396
* https://msdn.microsoft.com/en-us/library/dd340549.aspx?f=255&MSPPError=-2147217396
* https://msdn.microsoft.com/en-us/library/dd358082.aspx?f=255&MSPPError=-2147217396
[...]

PR #575, opened 2017-09-15
Microsoft SQL Server bcp import/export

AlexHagerman:
Added bcp functionality for odo within the framework of the other copy to/from flat file loaders. Tried to cover kwarg support for items like delimiter and newline settings. Using popen so that any error can be captured and populated. OSError will provide an error if bcp is not installed. One concern I have is building the Server.Database.Schema information. Currently I did this by slicing up the connection string that is given to sqlalchemy into a dict because I didn't see the server/db information in the element. Any feedback on a better way to accomplish retrieving that data that could be more stable would be appreciated.

### DIFF
--- a/odo/backends/sql.py
+++ b/odo/backends/sql.py
@@ -915,6 +915,28 @@ def compile_copy_to_csv_sqlite(element, compiler, **kwargs):
     # This will be a no-op since we're doing the write during the compile
     return ''
 
+@compiles(CopyToCSV, 'mssql')
+def compile_copy_to_csv_mssql(element, compiler, **kwargs):
+    selectable = element.element
+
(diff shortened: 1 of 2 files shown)


majidaldo:
@AlexHagerman do you have tests?

AlexHagerman:
I don't have any test. At the time I wrote this I wasn't as familiar with Python test frameworks. I've since made some progress if this is something anybody wants I can look into what mocks or fixtures would help facilitate testing bcp with Pytest. Since bcp is from a flat file I'd be open to suggestions on any fixtures that might make sense. Also not sure what SQL Server integration testing there is in the travis build if any.

nickolay:
There's some discussion of bulk-loading in https://github.com/blaze/odo/issues/571 and another attempt at adding bcp functionality among other things in https://github.com/blaze/odo/pull/525

Issue #567, opened 2017-08-21
BUG: mssql.TIMESTAMP -> bytes_ breaks discover


thequackdaddy:
Hello,

I noticed from [this commit](https://github.com/blaze/odo/commit/5c8232bf7eb73cca98e32cff3acc4fb18f4dcf18#diff-51da4c3b85f1cc5241513828c9882880R104) that I think we accidentally broke datetime discovery in blaze. 

Just try this code after this commit:

```python
import sqlalchemy as sa
from odo import discover
columns = [sa.Column('amount', sa.REAL),
           sa.Column('ds', sa.TIMESTAMP)]
dta = sa.Table('t', sa.MetaData(), *columns)
discover(dta)
```

this returns `dshape("var * {amount: ?float32, ds: ?bytes}")`. Note that 'ds' is now `?bytes`, not `timestamp`... 

(This breaks unittests in blaze too... trying to get those back)

@postelrich @dhirschfeld @llllllllll

PR #589, opened 2017-10-13
BUG: Fix discover for sa.TIMESTAMP.

ehebert:
`discover` was returning a bytes dtype for the `sa.TIMESTAMP` instead of
`datetime_`.

That bug was because the SQLAlchemy dialect for MSSQL currently directly imports
`sa.TIMESTAMP`, causing a collision in `odo.backends.sql.revtypes`.

Fix by 1) creating a subclass of `mssql.TIMESTAMP` to use as the key in
`revtypes` so that it does not overwrite `sa.TIMESTAMP` 2) assign that subclass
to the mssql dialect's `'TIMESTAMP'` (using `ischema_names`), so that the
subclass will be returned by the type engine instead of `mssql.TIMESTAMP`.

The added test for the (`sa.TIMESTAMP`, `datetime_`) without the fix applied
would result in this error:
```
a = Bytes(), b = DateTime(tz=None), path = ('.measure', "['ts']", '.ty')
kwargs = {'check_dim': True, 'check_record_order': True}

    @assert_dshape_equal.register(object, object)
    def _base_case(a, b, path=None, **kwargs):
>       assert a == b, '%s != %s\n%s' % (a, b, _fmt_path(path))
E       AssertionError: bytes != datetime
E       path: _.measure['value'].ty
```

This patch includes checks for the other types besides `sa.TIMESTAMP` for better
protection against the general case of failures when using `revtypes` to map
SQLAlchemy types to dtypes.

Those extra cases exposed an issue with `sa.Float(precision=24)`.
That case is commented out to keep the fix of this patch on the `sa.TIMESTAMP` mapping.
(I would have used `pytest.param` to mark it xfail, but pytest for this project needs to be
upgraded first.)

See:
https://github.com/blaze/odo/issues/567
https://github.com/blaze/odo/pull/568
https://bitbucket.org/zzzeek/sqlalchemy/issues/4092/type-problem-with-mssqltimestamp
https://github.com/blaze/blaze/pull/1656

### DIFF
--- a/odo/backends/tests/test_sql.py
+++ b/odo/backends/tests/test_sql.py
@@ -6,6 +6,7 @@ pytest.importorskip('sqlalchemy')
 import os
 from decimal import Decimal
 from functools import partial
+from textwrap import dedent
 
 import datashape
 from datashape import (
@@ -304,7 +305,16 @@ def test_discover_oracle_intervals(freq):
         (sa.types.NullType, string),
         (sa.REAL, float32),
         (sa.Float, float64),
-        # sa.Float(precision=24), float32 (reason="Currently returns float64")
+        pytest.param(
+            sa.Float(precision=24), float32,
+            marks=pytest.mark.xfail(
+                reason=dedent("""
+                Currently returns float64.
+                Unique instances of sa.Float(precision=24) do not equate to each other, which prevents
+                discover_typeengine from correctly keying into sql.revparses.
+                """).strip()
+            )
+        ),
         (sa.Float(precision=53), float64),
     ),
 )
--- a/odo/backends/tests/test_sql.py
+++ b/odo/backends/tests/test_sql.py
@@ -8,7 +8,20 @@ from decimal import Decimal
 from functools import partial
 
 import datashape
-from datashape import discover, dshape, float32, float64, Option
+from datashape import (
+    date_,
+    datetime_,
+    discover,
+    dshape,
+    int_,
+    int64,
+    float32,
+    float64,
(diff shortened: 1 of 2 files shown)


odo/backends/tests/test_sql.py
```diff
@@ -279,6 +292,28 @@ def test_discover_oracle_intervals(freq):
...
+        (sa.TIMESTAMP, datetime_),
+        (sa.FLOAT, float64),
+        (sa.DATE, date_),
+        (sa.BIGINT, int64),
+        (sa.INTEGER, int_),
+        (sa.BIGINT, int64),
+        (sa.types.NullType, string),
+        (sa.REAL, float32),
+        (sa.Float, float64),
+        # sa.Float(precision=24), float32 (reason="Currently returns float64")
```
llllllllll: looks like this should be a skip

llllllllll:
one comment but lgtm

odo/backends/tests/test_sql.py
ehebert: Yes, now that https://github.com/blaze/odo/pull/590 is merged, I'll use `pytest.param` to mark this as an `xfail`.

ehebert: https://github.com/blaze/odo/pull/589/commits/8e9f3c9d85ab2850bcb77870538bb7b493d1abc2

ehebert:
@llllllllll, when you merge this, could you force a rebuild of blaze's Travis suite?

PR #592, opened 2017-10-17
BUG: Fix reflection Postgres double precision columns.

ehebert:
A double precision column in a Postrges table, whose dshape specified a dtype of
`float64` for that column, would return a decimal dtype.

Fix by defining a group of types which are 'precision' types, and check for
membership in that before attempting to lookup the typeengine typ in
`blaze.backends.sql.revtypes`

Without the fix applied the newly added test fails with the following assertion
error, when comparing the `discover` of the `bind` with the expected values:
```
    @assert_dshape_equal.register(object, object)
    def _base_case(a, b, path=None, **kwargs):
>       assert a == b, '%s != %s\n%s' % (a, b, _fmt_path(path))
E       AssertionError: decimal[precision=53, scale=None] != float64
E       path: _.measure['a']
```

Also, added `sa.Float` as a precision type, which fixes
`sa.Float(precision=24)`; this allows removal of the `xfail` marking in
`odo/backends/test/test_sql::test_types`.

### DIFF
--- a/odo/backends/sql.py
+++ b/odo/backends/sql.py
@@ -21,7 +21,7 @@ from sqlalchemy import inspect
 from sqlalchemy.ext.compiler import compiles
 from sqlalchemy import event
 from sqlalchemy.schema import CreateSchema
-from sqlalchemy.dialects import mssql
+from sqlalchemy.dialects import mssql, postgresql
 
 from multipledispatch import MDNotImplementedError
 
@@ -107,8 +107,6 @@ revtypes.update({
     sa.types.NullType: string,
     sa.REAL: float32,
     sa.Float: float64,
-    sa.Float(precision=24): float32,
-    sa.Float(precision=53): float64,
     mssql.BIT: datashape.bool_,
     mssql.DATETIMEOFFSET: string,
     mssql.MONEY: float64,
@@ -118,6 +116,29 @@ revtypes.update({
     # It is instead just a binary(8) value with no relation to dates or times
     MSSQLTimestamp: bytes_,
 })
+
+# Types which can be specified on precision.
+# These are checked before checking membership in revtypes, because:
+# 1) An instance of a precision type does not equal another instancec with
+# the same precision.
+# (DOUBLE_PRECISION(precision=53) != DOUBLE_PRECISION(precision=53)
+# 2) Precision types can be a instance of a type in revtypes.
+# isinstance(sa.Float(precision=53), sa.Float)
+precision_types = {
+    sa.Float,
+    postgresql.base.DOUBLE_PRECISION
+}
+
+# Maps the value of a precision types `precision` attribute to the desired
+# dtype.
+# e.g. the value returned by
+# `postgresql.base.DOUBLE_PRECISION(precision=53).precision`
+# maps to `float64`.
+precision_to_dtype = {
+    24: float32,
+    53: float64
+}
+
 
 # interval types are special cased in discover_typeengine so remove them from
 # revtypes
@@ -199,6 +220,8 @@ def discover_typeengine(typ):
                              'second_precision=%d, day_precision=%d' %
                              (typ.second_precision, typ.day_precision))
         return datashape.TimeDelta(unit=units)
+    if type(typ) in precision_types and typ.precision in precision_to_dtype:
+        return precision_to_dtype[typ.precision]
     if typ in revtypes:
         return dshape(revtypes[typ])[0]
     if type(typ) in revtypes:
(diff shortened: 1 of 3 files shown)


odo/backends/sql.py
```diff
@@ -119,6 +117,29 @@ class MSSQLTimestamp(mssql.TIMESTAMP):
     MSSQLTimestamp: bytes_,
 })
 
+# Types which can be specified on precision.
+# These are checked before checking member ship in revtypes, because:
```
llllllllll: typo: membership

odo/backends/sql.py
```diff
@@ -199,6 +220,8 @@ def discover_typeengine(typ):
                              'second_precision=%d, day_precision=%d' %
                              (typ.second_precision, typ.day_precision))
         return datashape.TimeDelta(unit=units)
+    if type(typ) in precision_types and typ.precision in precision_to_dtype:
```
llllllllll: what happens when the first is true but the second is not?

odo/backends/tests/test_sql.py
```diff
@@ -305,24 +305,15 @@ def test_discover_oracle_intervals(freq):
...
-            sa.Float(precision=24), float32,
-            marks=pytest.mark.xfail(
-                reason=dedent("""
-                Currently returns float64.
-                Unique instances of sa.Float(precision=24) do not equate to each other, which prevents
-                discover_typeengine from correctly keying into sql.revparses.
-                """).strip()
-            )
-        ),
+        (sa.Float(precision=24), float32),
```
llllllllll: should we also test precision=53 and something that is not one of the hard coded numbers?

ehebert: The `precision=53` case is defined in the next line.
Good idea about adding a test for another number.

odo/backends/sql.py
ehebert: A `sa.Float(precision=42)` (where 42 is not a mapped precision value) would be treated as a `sa.Float` by the `if type(typ) in revtypes` predicate.

A postgres `DOUBLE_PRECISION` would return a `Decimal(precision=42)`, having gone through the logic which this patch circumvents when the precision is 53.

I'm not sure what should be the correct logic when there is a precision but it is not 24 or 53.
A couple options:
1) If the precision is not None is not mapped, raise an exception (Something like `UnexpectedColumnPrecision`.)  If the precision is None, then use the mapping in `revtypes`.
e.g. `sa.Float(precision=42)` raises an exception; `sa.Float()` returns `float64` (The same dtype as `sa.Float`)
2) If the precision is not mapped, then just continue on with the previously defined dtype mapping logic . (This is the logic the PR currently implements.)
In this case, `sa.Float(precision=8)`, `sa.Float(precision=42)`, `sa.Float()` all return `float64`.

I lean towards raising the exception on unrecognized precisions.

odo/backends/tests/test_sql.py
ehebert: The expected case for `Float(precision=42)` would depend on the answer to the questions about behavior here, https://github.com/blaze/odo/pull/592#discussion_r145212373

odo/backends/sql.py
llllllllll: I agree, raising an exception seems like the best thing to do for now.

llllllllll: Actually, what is wrong with returning 64? It will be able to hold the value.

ehebert: How about instead of mapping just 24 and 53, instead mapping based on ranges of values:
- `1-24` to `float32`
- `25-53` to `float64`
- `>=54` raises an exception.

This would follow the behavior of Postgres.
From https://www.postgresql.org/docs/9.5/static/datatype-numeric.html#DATATYPE-NUMERIC-DECIMAL,
```
PostgreSQL also supports the SQL-standard notations float and float(p) for specifying inexact numeric types. Here, p specifies the minimum acceptable precision in binary digits. PostgreSQL accepts float(1) to float(24) as selecting the real type, while float(25) to float(53) select double precision. Values of p outside the allowed range draw an error. float with no precision specified is taken to mean double precision
```

ehebert: Thinking about the `>=54`, since that is not a valid value on the database side, that would not be a precision returned during reflection, so just the supported ranges would need to be handled.

llllllllll: I think it doesn't hurt to guard against that for now, I like the range mappings.

ehebert: https://github.com/blaze/odo/pull/592/commits/a3785c6aaf8dcd98237bca27e5ff1079a910f480

llllllllll:
Thanks for this!

PR #594, opened 2017-10-23
BUG: allow other connectables to be passed as the bind param

llllllllll:
Allows you to do stuff like:

```python
with engine.begin() as conn:
    odo(..., bind=conn)
    odo(..., bind=conn)

and have both calls run in the same transaction.

### DIFF
--- a/odo/backends/sql.py
+++ b/odo/backends/sql.py
@@ -181,7 +181,7 @@ def getbind(t, bind):
     if bind is None:
         return t.bind
 
-    if isinstance(bind, sa.engine.base.Engine):
+    if isinstance(bind, sa.engine.interfaces.Connectable):
         return bind
     return create_engine(bind)
 
@@ -538,9 +538,9 @@ def select_to_base(sel, dshape=None, bind=None, **kwargs):
 @append.register(sa.Table, Iterator)
 def append_iterator_to_table(t, rows, dshape=None, bind=None, **kwargs):
     assert not isinstance(t, type)
-    engine = getbind(t, bind)
-    if not t.exists(bind=engine):
-        t.create(bind=engine)
+    bind = getbind(t, bind)
+    if not t.exists(bind=bind):
+        t.create(bind=bind)
     rows = iter(rows)
 
     # We see if the sequence is of tuples or dicts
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by llllllllll** [comment]
@kglowinski

File: odo/backends/sql.py [hunk 1]
```diff
@@ -181,7 +181,7 @@ def getbind(t, bind):
     if bind is None:
         return t.bind
 
-    if isinstance(bind, sa.engine.base.Engine):
+    if isinstance(bind, sa.engine.interfaces.Connectable):
```
kglowinski: Are there other connectables in here that should be explicitly wired?

odo/backends/sql.py
```diff
@@ -607,8 +607,7 @@ def append_select_statement_to_sql_Table(t, o, bind=None, **kwargs):
 
     query = t.insert().from_select(o.columns.keys(), o)
 
-    with bind.connect() as conn:
-        conn.execute(query)
+    bind.execute(query)
```
kglowinski: Why does this no longer need to be in the begin context?

llllllllll: each thing executed is in its own mini-transaction unless explicitly in a transaction already, this was redundant.

odo/backends/sql.py
```diff
@@ -538,9 +538,9 @@ def select_to_base(sel, dshape=None, bind=None, **kwargs):
 @append.register(sa.Table, Iterator)
 def append_iterator_to_table(t, rows, dshape=None, bind=None, **kwargs):
     assert not isinstance(t, type)
-    engine = getbind(t, bind)
-    if not t.exists(bind=engine):
-        t.create(bind=engine)
+    bind = getbind(t, bind)
```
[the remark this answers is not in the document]
llllllllll: unless it is not a connectable in which case it will convert it into an engine. This allows you to pass a db uri instead.

odo/backends/sql.py
llllllllll: connectable is the interface, so anything that works here should be fine.

llllllllll:
I am not worried about the mysql tests right now, it is a known issue.

Issue #556, opened 2017-07-10
test_convert_dataframe_to_cds fails


dhirschfeld:
On win64/py36, bokeh 0.12.6, pandas 0.20.2 I'm seeing the below error:
```
________________________ test_convert_dataframe_to_cds ________________________

    def test_convert_dataframe_to_cds():
        cds = convert(ColumnDataSource, df)
>       assert cds.data == {'name': ['Alice', 'Bob', 'Charlie'],
                            'balance': [100, 200, 300]}

odo\backends\tests\test_bokeh.py:18:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = 0    True
1    True
2    True
Name: balance, dtype: bool

    def __nonzero__(self):
        raise ValueError("The truth value of a {0} is ambiguous. "
                         "Use a.empty, a.bool(), a.item(), a.any() or a.all()."
>                        .format(self.__class__.__name__))
E       ValueError: The truth value of a Series is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().
```
The test does seem to work on Travis though. I'm assuming it must be testing against an older version of bokeh or pandas?

If I inspect `cds.data` this is what I'm seeing:
```python
In [49]: from odo.backends.bokeh import convert, pd, ColumnDataSource

In [50]: df = pd.DataFrame([[100, 'Alice'],
    ...:                    [200, 'Bob'],
    ...:                    [300, 'Charlie']],
    ...:                   columns=['balance', 'name'])

In [51]: cds = convert(ColumnDataSource, df)

In [52]: cds.data
Out[52]: 
{'balance': 0    100
 1    200
 2    300
 Name: balance, dtype: int64, 'name': 0      Alice
 1        Bob
 2    Charlie
 Name: name, dtype: object}
```

PR #596, opened 2017-11-02
Fix detection of integral dtypes

dhirschfeld:
**TL;DR;**
`if ty in datashape.integral` doesn't detect `?decimal[precision=1, scale=0]` as integral but `ty.to_numpy_dtype()` returns `np.int32` which then blows up on nulls (None)

**Also Fixes:**
- #556
- #597

### DIFF
--- a/odo/backends/sql.py
+++ b/odo/backends/sql.py
@@ -88,7 +88,7 @@ revtypes = dict(map(reversed, types.items()))
 # mssql.TIMESTAMP and sa.TIMESTAMP.
 # At the time of this writing, (mssql.TIMESTAMP == sa.TIMESTAMP) is True,
 # which causes a collision when defining the revtypes mappings.
-# 
+#
 # See:
 # https://bitbucket.org/zzzeek/sqlalchemy/issues/4092/type-problem-with-mssqltimestamp
 class MSSQLTimestamp(mssql.TIMESTAMP):
@@ -207,7 +207,7 @@ def batch(sel, chunksize=10000, bind=None):
                     yield rows
                 else:
                     return
-    
+
     columns = [col.name for col in sel.columns]
     iterator = rowiterator(sel)
     return columns, concat(iterator)
@@ -794,13 +794,14 @@ def select_or_selectable_to_frame(el, bind=None, dshape=None, **kwargs):
     for field, dtype in fields:
         if isinstance(dtype, Option):
             ty = dtype.ty
-            if ty in datashape.integral:
-                dtypes[field] = 'float64'
+            try:
+                dtypes[field] = ty.to_numpy_dtype()
+            except TypeError:
+                dtypes[field] = np.dtype(object)
             else:
-                try:
-                    dtypes[field] = ty.to_numpy_dtype()
-                except TypeError:
-                    dtypes[field] = np.dtype(object)
+                if np.issubdtype(dtypes[field], np.integer):
+                    # cast nullable ints to float64 so NaN can be used for nulls
+                    dtypes[field] = np.float64
         else:
             try:
                 dtypes[field] = dtype.to_numpy_dtype()
--- a/odo/backends/tests/test_bokeh.py
+++ b/odo/backends/tests/test_bokeh.py
@@ -15,9 +15,8 @@ df = pd.DataFrame([[100, 'Alice'],
 
 def test_convert_dataframe_to_cds():
     cds = convert(ColumnDataSource, df)
-    assert cds.data == {'name': ['Alice', 'Bob', 'Charlie'],
-                        'balance': [100, 200, 300]}
-
+    assert list(cds.data['name']) == ['Alice', 'Bob', 'Charlie']
+    assert list(cds.data['balance']) == [100, 200, 300]
     df2 = convert(pd.DataFrame, cds)
     assert isinstance(df2, pd.DataFrame)
 
--- a/odo/backends/pandas.py
+++ b/odo/backends/pandas.py
@@ -14,7 +14,11 @@ from ..convert import convert
 
 
 possibly_missing = frozenset({string, datetime_})
-categorical = type(pd.Categorical.dtype)
+try:
+    from pandas.api.types import CategoricalDtype as categorical
+except ImportError:
+    categorical = type(pd.Categorical.dtype)
+    assert categorical is not property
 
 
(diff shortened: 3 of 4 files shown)


dhirschfeld:
**Repro**:
```python
import blaze as bz
from numpy.random import randn
from odo import odo
import pandas as pd
import sqlalchemy as sa


engine = sa.create_engine('sqlite://')
metadata = sa.MetaData(bind=engine)


T = sa.Table(
    'Demo',
    metadata,
    sa.Column('pkid', sa.Integer, primary_key=True),
    sa.Column('value', sa.DECIMAL(precision=1, scale=0, asdecimal=False), nullable=True),
)

metadata.create_all()

data = pd.DataFrame(list(zip(range(10), randn(10) > 0)), columns=['pkid','value'])
data.loc[5, 'value'] = None
records = data.to_dict(orient='records')

with engine.connect() as conn:
    conn.execute(T.insert(), records)
```
```python
In [2]: T.select().execute().fetchall()
Out[2]: 
[(0, 0),
 (1, 1),
 (2, 1),
 (3, 0),
 (4, 0),
 (5, None),
 (6, 0),
 (7, 0),
 (8, 1),
 (9, 0)]

In [4]: bz.discover(T)
Out[4]: dshape("var * {pkid: int32, value: ?decimal[precision=1, scale=0]}")

In [5]: odo(T, pd.DataFrame)
Traceback (most recent call last):

  File "<ipython-input-5-4015a811a204>", line 1, in <module>
    odo(T, pd.DataFrame)

  File "C:\Miniconda3\lib\site-packages\odo\odo.py", line 91, in odo
    return into(target, source, **kwargs)

  File "C:\Miniconda3\lib\site-packages\multipledispatch\dispatcher.py", line 164, in __call__
    return func(*args, **kwargs)

  File "C:\Miniconda3\lib\site-packages\odo\into.py", line 43, in wrapped
    return f(*args, **kwargs)

  File "C:\Miniconda3\lib\site-packages\odo\into.py", line 53, in into_type
    return convert(a, b, dshape=dshape, **kwargs)

  File "C:\Miniconda3\lib\site-packages\odo\core.py", line 83, in __call__
    return _transform(self.graph, *args, **kwargs)

  File "C:\Miniconda3\lib\site-packages\odo\core.py", line 106, in _transform
    x = f(x, excluded_edges=excluded_edges, **kwargs)

  File "C:\Miniconda3\lib\site-packages\odo\backends\sql.py", line 757, in select_or_selectable_to_frame
    dtype=[(str(c), dtypes[c]) for c in columns]))

TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

dhirschfeld:
```
<snip>
odo/backends/tests/test_sql.py::test_integer_detection PASSED
<snip>
======= 457 passed, 45 skipped, 8 xfailed, 5 warnings in 27.74 seconds ========
```

dhirschfeld:
@llllllllll, @ehebert, @postelrich - it would be good if someone could hit the merge button on this one as it fixes a number of issues, including compatibility with the latest pandas. As mentioned above, the travis failure seems unrelated.

necaris:
LGTM

dhirschfeld:
Could this PR please be merged - it fixes a number of bugs which are affecting my use of odo in a production system.

@llllllllll - I understand that there's no full-time maintainer for odo but if no one else has time to review and merge bug-fix PRs perhaps I could be given commit rights myself to do so?

ehebert:
@dhirschfeld to help expedite a merge I can offer to review your work soon. However, I currently also do not have commit rights.

dhirschfeld:
@llllllllll - any chance you could merge this? Happy to make any suggested changes...

odo/backends/sql.py
```diff
@@ -794,13 +794,14 @@ def select_or_selectable_to_frame(el, bind=None, dshape=None, **kwargs):
...
+            try:
+                dtypes[field] = ty.to_numpy_dtype()
+            except TypeError:
+                dtypes[field] = np.dtype(object)
             else:
-                try:
-                    dtypes[field] = ty.to_numpy_dtype()
-                except TypeError:
-                    dtypes[field] = np.dtype(object)
+                if np.issubdtype(dtypes[field], np.integer):
```
postelrich: Does this else-if branch leave the possibility of no type being converted for that field?

dhirschfeld: It's a try/except/else branch so the `if np.issubdtype(dtypes[field], np.integer)` block runs only if the try has already *succeeded*:
```python
try:
    dtypes[field] = ty.to_numpy_dtype()
except TypeError:
    dtypes[field] = np.dtype(object)
else:
    # If `ty.to_numpy_dtype()` results in an integer type override `dtypes[field]` to be float64
    # We do this because we already know `dtype` is an `Option` type and pandas uses NaN
    # as a null placeholder requiring that the type be floating point rather than integral
    if np.issubdtype(dtypes[field], np.integer):
        # cast nullable ints to float64 so NaN can be used for nulls
        dtypes[field] = np.float64
```

dhirschfeld: 26e300e5219cbc667831b54ba55254a58fae75ba is a self-contained repro of the underlying issue

ehebert:
@dhirschfeld I've blocked out some time today to work on merging this.

odo/backends/tests/test_bokeh.py
```diff
@@ -15,9 +15,8 @@
 
 def test_convert_dataframe_to_cds():
     cds = convert(ColumnDataSource, df)
-    assert cds.data == {'name': ['Alice', 'Bob', 'Charlie'],
-                        'balance': [100, 200, 300]}
-
+    assert list(cds.data['name']) == ['Alice', 'Bob', 'Charlie']
```
ehebert: This is orthogonal to the integral dtype change, correct?
(It's ok to leave in, especially since we want to  move towards getting tests passing. Just wanted to make sure that I wasn't missing some relation to the integral dtype detection.)

dhirschfeld: Yes it prob should have be in a separate "Fix the tests" PR. I fixed 2 tests trying to get the CI green on this PR before I gave up. I'm happy to look into the other failures and see what I can do there to finish the job.

If I get around to it I'll make sure to open an issue first so people know I'm looking into it and we don't duplicate any work.

Issue #579, opened 2017-09-22
Odo 0.5.0 is incompatible with networkx 2.0 on Windows


mcquin:
`odo.odo` fails on Windows when inserting a CSV file into a (sqlite) database due to networkx 2.0.

* Windows 10
* Python 2.7.14

odo (and networkx) installed/upgraded via command `pip install --upgrade odo`.

[test file](https://gist.githubusercontent.com/anonymous/ef02e403b8af9fcfaaea3d6af9459b31/raw/2e1844cb37766345948fe4e561aa8a14f32b2670/pets.csv)

odo 0.5.0 with networkx 2.0:
```
Python 2.7.14 (v2.7.14:84471935ed, Sep 16 2017, 20:19:30) [MSC v.1500 32 bit (Intel)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import odo
C:\Python27\lib\site-packages\odo\backends\pandas.py:94: FutureWarning: pandas.tslib is deprecated and will be removed i
n a future version.
You can access NaTType as type(pandas.NaT)
  @convert.register((pd.Timestamp, pd.Timedelta), (pd.tslib.NaTType, type(None)))
>>> odo.odo("pets.csv", "sqlite:///foo.db::Pets")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Python27\lib\site-packages\odo\odo.py", line 91, in odo
    return into(target, source, **kwargs)
  File "C:\Python27\lib\site-packages\multipledispatch\dispatcher.py", line 164, in __call__
    return func(*args, **kwargs)
  File "C:\Python27\lib\site-packages\odo\into.py", line 43, in wrapped
    return f(*args, **kwargs)
  File "C:\Python27\lib\site-packages\odo\into.py", line 149, in into_string_string
    return into(a, resource(b, **kwargs), **kwargs)
  File "C:\Python27\lib\site-packages\multipledispatch\dispatcher.py", line 164, in __call__
    return func(*args, **kwargs)
  File "C:\Python27\lib\site-packages\odo\into.py", line 43, in wrapped
    return f(*args, **kwargs)
  File "C:\Python27\lib\site-packages\odo\into.py", line 143, in into_string
    return into(a, b, dshape=dshape, **kwargs)
  File "C:\Python27\lib\site-packages\multipledispatch\dispatcher.py", line 164, in __call__
    return func(*args, **kwargs)
[...]

PR #601, opened 2018-01-10
Compatibility Networkx 2

michelorengo:
Fix for #579


odo/core.py
```diff
@@ -169,10 +169,14 @@ def path(graph, source, target, excluded_edges=None, ooc_types=ooc_types):
...
-        edge = graph.edge
+        #https: // github.com / blaze / odo / issues / 579
+        #https://github.com/blaze/odo/issues/588
+        #node can be retrieved from graph via dict-like access
+        #edge = graph.edge
 
         def path_part(src, tgt):
-            node = edge[src][tgt]
+            #node = edge[src][tgt]
+            node = graph[src][tgt]
```
postelrich: Can you remove the comments? A git blame on github should suffice in referencing the change.

michelorengo: It is done.

Issue #585, opened 2017-10-02
CSV to sqlite fails with "TypeError: can only concatenate list (not "TextFileReader") to list"


makmanalp:
Code run: `odo("./test.csv", "sqlite:///database.db::country_partner_year")`.

The offending line is at: https://github.com/blaze/odo/blob/39bf5770935c7cb50ec400aec378857379ba88bd/odo/backends/csv.py#L371

Pandas (v0.20.3) seems to return a TextFileReader object when setting `chunksize=blah`, and so first is of type list[DataFrame], and rest is of type TextFileReader, and adding them doesn't make any sense - this seems to be a job for `pd.concat()` or something similar, perhaps `itertools.chain` depending on what exactly is happening after that function.

Full backtrace:

<details>
/nfs/projects_nobackup/c/cidgrowlab/Mali/intl-atlas-api/env/lib/python3.4/site-packages/odo/odo.py in odo(source, target, **kwargs)
     89     odo.append.append      - Add things onto existing things
     90     """
---> 91     return into(target, source, **kwargs)

/nfs/projects_nobackup/c/cidgrowlab/Mali/intl-atlas-api/env/lib/python3.4/site-packages/multipledispatch/dispatcher.py in __call__(self, *args, **kwargs)
    162             self._cache[types] = func
    163         try:
--> 164             return func(*args, **kwargs)
    165 
    166         except MDNotImplementedError:

/nfs/projects_nobackup/c/cidgrowlab/Mali/intl-atlas-api/env/lib/python3.4/site-packages/odo/into.py in wrapped(*args, **kwargs)
     41             raise TypeError('dshape argument is not an instance of DataShape')
     42         kwargs['dshape'] = dshape
---> 43         return f(*args, **kwargs)
     44     return wrapped
     45 

/nfs/projects_nobackup/c/cidgrowlab/Mali/intl-atlas-api/env/lib/python3.4/site-packages/odo/into.py in into_string_string(a, b, **kwargs)
    147 @validate
    148 def into_string_string(a, b, **kwargs):
--> 149     return into(a, resource(b, **kwargs), **kwargs)
    150 
    151 

/nfs/projects_nobackup/c/cidgrowlab/Mali/intl-atlas-api/env/lib/python3.4/site-packages/multipledispatch/dispatcher.py in __call__(self, *args, **kwargs)
    162             self._cache[types] = func
    163         try:
--> 164             return func(*args, **kwargs)
    165 
    166         except MDNotImplementedError:

[...]

PR #605, opened 2018-02-06
fix large csv reading #585

majidaldo:
use file path and chain df iterators

### DIFF
--- a/odo/backends/csv.py
+++ b/odo/backends/csv.py
@@ -33,6 +33,8 @@ from ..temp import Temp
 from ..numpy_dtype import dshape_to_pandas
 from .pandas import coerce_datetimes
 from functools import partial
+from itertools import chain
+
 
 dialect_terms = '''delimiter doublequote escapechar lineterminator quotechar
 quoting skipinitialspace strict'''.split()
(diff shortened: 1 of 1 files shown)


postelrich:
What specifically is this fixing?

majidaldo:
https://github.com/blaze/odo/issues/585

odo/backends/csv.py
```diff
@@ -368,7 +369,7 @@ def CSV_to_chunks_of_dataframes(c, chunksize=2 ** 20, **kwargs):
     else:
         rest = []
 
-    data = [first] + rest
+    data = chain([first], rest)
```
brandonwillard: `toolz.chain` should be able to do this, no?  It's already imported, too.

majidaldo: there is no toolz.chain. similar functions in toolz just import itertools.
toolz source:
```
def concat(seqs):
    """ Concatenate zero or more iterables, any of which may be infinite.

    An infinite sequence will prevent the rest of the arguments from
    being included.

    We use chain.from_iterable rather than ``chain(*seqs)`` so that seqs
    can be a generator.

    >>> list(concat([[], [1], [2, 3]]))
    [1, 2, 3]

    See also:
        itertools.chain.from_iterable  equivalent
    """
    return itertools.chain.from_iterable(seqs)



def concatv(*seqs):
    """ Variadic version of concat

    >>> list(concatv([], ["a"], ["b", "c"]))
    ['a', 'b', 'c']

    See also:
        itertools.chain
    """
    return concat(seqs)

```

brandonwillard: Sorry, typo; I meant `toolz.concat`.

majidaldo: do you want to change it?

TBusen:
is this still being worked?

majidaldo:
no. it works.

guillochon:
Any way to merge this soon? Currently preventing me from using odo.

TBusen:
Same here, I can’t use odo

PR #606, opened 2018-02-15
Python 3.6

postelrich:
This PR adds support for python 3.6 as well as updating dependencies to more current versions.


.travis.yml
```diff
@@ -40,6 +40,25 @@ install:
   - conda config --set always_yes yes --set changeps1 no
   - conda update conda
 
+<<<<<<< HEAD
```
dhirschfeld: Conflict marker?

postelrich: 👍 , that might be why its trying to do a ruby build

PR #607, opened 2018-02-17
Replace a default string value with `None`.

yuvallanger:
The previous version was redundant.


yuvallanger:
On second thought, I now realize it might be used as a Python 2 compatible way to type annotate the functions? Why do it this way?

PR #616, opened 2018-09-22
Make odo compatible with NetworkX >=2.0

### LINKED ISSUE #615: AttributeError: 'DiGraph' object has no attribute 'edge'  - when trying to use odo
I'm trying to use odo  and i'm getting the above error message. 
Down grading networkx seems to be one suggest solution, but is proving to be a hassle? 
When will odo work with networkx 2.1? 
Or is there another cause/ fix. 

thanx

groutr:
NetworkX changed its API for accessing edges.  This PR makes odo compatible with NetworkX 1.0/2.0 by adding a compatibility function for each version of NetworkX.

Fixes #615

### DIFF
--- a/etc/ci-install.sh
+++ b/etc/ci-install.sh
@@ -1,3 +1,5 @@
+set -x
+
 # Install dependencies
 # Use conda **ONLY** for numpy and pandas (if not pulling from master), this
 # speeds up the builds a lot. Use the normal pip install for the rest.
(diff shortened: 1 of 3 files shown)


soerendip:
Thanks for working on this.

dhirschfeld:
It seems pretty clean and self-contained so I don't see any harm in merging it.

bhipple:
Any chance we could get a new release with this?  When one is cut we can unwind some of the distribution patches for this in NixPkgs and simplify our packaging config, which is always nice!

